Issue #288, opened 2016-11-15
api: add upsert entrypoint for route updates


ccirello:
Currently we can only issue POST (create), DELETE and and GET to Routes. So whenever faced with the need to update a route, we are forced to do a delete-then-create for the routes; triggering service disruption during interfaces. 

fnctl takes this approach for now - and fixing this issue must also change its source code.

It relates to https://github.com/iron-io/functions/pull/259 and its related issues.

ccirello:
cc @pedronasser

ccirello:
cc @denismakogon

ccirello:
```
./fnctl routes config set myapp hello key1 value1
./fnctl routes config set myapp hello key2 value2
```
the key2/value2 pair will not be there.

PR #295, opened 2016-11-15
Fix route update

pedronasser:
- Added EnsureImage to route update

Relates to #288


treeder:
@ccirello can you review

ccirello:
I am testing it...

ccirello:
It might have fixed something, but definitely it did not help with https://github.com/iron-io/functions/issues/288

ccirello:
This did not help with #288 - but I think I see the point of this change. In the end, this change would be necessary to ensure that route updates work. So partial route updates, ie the ones that are not changing the image, still go through.

Issue #77, opened 2016-08-29
Dynamic routes: Can we optimize the route matching rather than pulling up all routes for an app?


treeder:
This is tricky because of this: https://github.com/iron-io/functions/issues/71

ie: https://github.com/iron-io/functions/pull/74/files#diff-e180f07f5c320b82004c2bb065eedef0R99

The problem is we'd like to query for a single route, but let's say the route is defined as: 

```
/blogs/:blog_id/comments/:comment_id
```

And a request comes in for:

```
/blogs/123/comments/456
```

How do we query the database for the route metadata with only that information?  Could query for `/blogs/123/comments/456`, which is the static default way. But that wouldn't match. We could try all variations, eg `/blogs/123/comments/*` and `/blogs/123/*/456`, and `/blogs/123/*/*`, and so on, but that wouldn't be very efficient. 

To summarize, the problem here is:

Given a path and only a path, eg: `/blogs/travisblog/comments/123` what should you query for?  Not knowing what routes exist before hand.

jmank88:
I came up with a scheme for postgres that at least pushes all the work to the server, so all the routes don't come over the wire each time, though I'm still not sure how to index it efficiently.

If the routes are stored with parameter names removed (Example: `/blogs/:blog_id/comments/:comment_id` -> `/blogs/:/comment/:`), then fairly simple regex queries can be used by matching on either the whole word or the colon character at each path component. Example: `/blogs/123/comments/456` -> `^/(blogs|:)/(123|:)/(comments|:)/(456|:)$`

```sql
CREATE TABLE routes (
  route TEXT,
  raw_route TEXT
);

-- Register routes
INSERT INTO routes (route, raw_route) VALUES 
('/blogs/:blog_id/comments/:comment_id', '/blogs/:/comments/:'),
('/blogs/:blog_id/comments', '/blogs/:/comments'),
('/blogs/:blog_id','/blogs/:'),
('/blogs','/blogs');

-- Query for '/blogs/123/comments/456'
SELECT route FROM routes WHERE raw_route ~ '^/(blogs|:)/(123|:)/(comments|:)/(456|:)$' LIMIT 1
```

In theory, the same general approach could translate to BoltDB as well - store by nameless route components in nested buckets, and then search for matches one component at a time.

jmank88:
I implemented a prefix-scanning solution for BoltDB: https://godoc.org/github.com/jmank88/nuts#hdr-Path_Prefix_Scans
[...]

PR #300, opened 2016-11-15
[Proposal] functions: restore dynamic routes

ccirello:
This PR reintroduces the support for dynamic route matching (the ones with `:` in parts of the address). Before, it would iterate throughout all routes of a given application, not matter what. Now, it takes a longer path serving each request, but allowing for early exists.

Step by step, each request is first checked against an LRU cache - which stores matchable routes, then incoming request is sought in the database, either dynamic or not. 

It adds some extra points of improvements. First, it primes the cache during start time. Thus helping LRU to be effective from the start. Second, the LRU size is limited to 20% of all routes of an application (Pareto rule: 20% of routes will supposedly do 80% of the workload). Third, it uses the singleflight pattern to easy the load on the database while loading the routes for dynamic route matching.

Obviously this is not a complete solution yet. Some important aspects are missing, for future PRs. They are: a) we need to find a way to iterate only through a small subset of the routes (check the code for proposals on how to do it in the TODO comment at api/server/runner.go); b) we need to find a way to impede DoS in the presence of a flood of certainly unknown routes requests (this is will be a tough problem).

Addresses partially #77


treeder:
Conversation: https://open-iron.slack.com/archives/functions/p1479513871000630

ccirello:
Proposal rejected.

PR #305, opened 2016-11-15
Client build fixes

treeder:


ccirello:
I wonder whether we couldn't convert this to Go sometime.

treeder:
Yes, we probably should at some point.

Issue #165, opened 2016-10-14
All runners could be sync and async to optimize resources


treeder:
Share same pool of runners for sync and async. 

So I was thinking about how this could work on a single binary/machine (instead of separate runners), and had this idea:

* a request comes in
* if sync, put in sync channel - implemented (partially?) here #92 
* if async, put in message queue

Then we just have a single go routine that does the following:

* read off channel, if task exists, fire up go routine to execute it, write the results to the HTTP response. If no sync tasks:
* read off message queue, if task exists, fire up go routine to execute it (probably nearly the exact same way as above, except we don't write any results

Or maybe those are two go routines, but the second one pauses until resources are available.

Prior convo here: https://github.com/iron-io/functions/issues/104#issuecomment-248947222 for details.

ccirello:
Done in #314

treeder:
I'm pretty sure this isn't done. Async is still totally separate from sync from what I can tell.

ccirello:
How sure are you sure? Because I am pretty confident that I have unified sync and async jobs to use the same workers.

ccirello:
Check https://github.com/iron-io/functions/blob/master/api/runner/async_runner.go#L130 -- as you can see, it talks to `tasks` channel - which is the same channel that takes in sync tasks. So they _were_ separated stuff, not anymore. Moving convo to slack.

treeder:
Ok, thanks for link.

Issue #69, opened 2016-08-24
Long(er) running containers for better performance aka Hot Containers


treeder:
Instead of one off execution, keep container running for a while and take requests.  

The first version executes a container per function call/request. Which is fine for many things, but we may want a way to run for longer periods of time to increase performance due to things like keeping db connections open, etc. 

This would probably require a more lambda like code format and language libraries. See https://github.com/iron-io/lambda for existing work on lambda style code format. 

I created a prototype of this years ago that routed traffic into IronWorker containers, should dig that out and look into that implementation.

ccirello:
See #214

ccirello:
https://github.com/iron-io/functions/pull/332 fixes this

Issue #10, opened 2016-07-19
Wait for capacity if no RAM left on server and log wait time as metric to logs


treeder:
Share memory checking functions from Titan. 

If queue of requests get too big (100?), start returning 503's. 

ALWAYS log the wait time, even if it's zero.

PR #314, opened 2016-11-16
functions: common concurrency stream for sync and async

ccirello:
This is a proposal that addresses concerns some aspects of https://github.com/iron-io/functions/issues/165  and https://github.com/iron-io/functions/issues/69.

1 - We have now a supervised goroutine that spawns workers for the incoming load. They serve any kind of request, sync or async (#165)

2 - Because runners start are separate from request life-cycle, we would be a third way closer to have hot reusable containers (#69)

Edit: 
***Implementation Details***

This PR introduces new goroutine responsible solely for executing tasks. It interfaces with the HTTP server and async modules through a channel, whose contents holds both the task information, and a channel to send the answer backer to the caller. https://github.com/iron-io/functions/blob/bounded-concurrency/api/server/runner.go#L183-L188 and https://github.com/iron-io/functions/blob/bounded-concurrency/api/runner/async_runner.go#L142-L145

In order to prevent resource starvation between competing sync and async tasks, we reserve a maxium of 50% of the memory for async tasks. This number is arbitrary and might be subject to changes later. https://github.com/iron-io/functions/blob/bounded-concurrency/api/runner/runner.go#L118-L123

In practice, when the memory allocated for async workers is maxed out, it keeps looping until memory is available: https://github.com/iron-io/functions/blob/bounded-concurrency/api/runner/async_runner.go#L117-L121

Also relevant is that `Server` now fathers handleRequest https://github.com/iron-io/functions/blob/bounded-concurrency/api/server/runner.go#L37 - with which it injects the tasks channel into the HTTP server.

### DIFF
--- a/api/runner/async_runner.go
+++ b/api/runner/async_runner.go
@@ -140,7 +140,7 @@ func startAsyncRunners(ctx context.Context, url string, tasks chan TaskRequest, 
 			go func() {
 				defer wg.Done()
 				tresp := make(chan TaskResponse)
-				treq := TaskRequest{Prio: Low, Ctx: ctx, Config: getCfg(task), Response: tresp}
+				treq := TaskRequest{Ctx: ctx, Config: getCfg(task), Response: tresp}
 				tasks <- treq
 				resp := <-treq.Response
 				// Process Task
--- a/api/runner/async_runner.go
+++ b/api/runner/async_runner.go
@@ -9,9 +9,8 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"time"
-
 	"sync"
+	"time"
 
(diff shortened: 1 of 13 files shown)


treeder:
Couple questions @ccirello:

* The work done in #10 should have made it bound to the max number of runners, with a small buffer (100 requests), then the rest get rejected. Unless that didn't get implemented like the issue description? cc @pedronasser 
* What's the point of resizing the number of runners?  There's always some fixed limit to the memory available right at the start. We can't go above it and there's no loss for keeping them running when below.

ccirello:
@treeder now that I have the information that the runner regulates itself, there's no need for max runners. I am going to remove this extra mechanism.

PR #322, opened 2016-11-19
functions: performance improvements - LRU & singleflight DB calls

ccirello:


api/server/runner.go
```diff
@@ -76,44 +76,68 @@ func (s *Server) handleRequest(c *gin.Context, enqueue models.Enqueue) {
...
+	resp, err := s.singleflight.do(
+		filter,
+		func() (interface{}, error) {
+			return Api.Datastore.GetRoutesByApp(filter.AppName, &filter)
+		},
+	)
+	return resp.([]*models.Route), err
+}
+
+func (s *Server) serve(c *gin.Context, log logrus.FieldLogger, appName string, found *models.Route, app *models.App, route, reqID string, payload io.Reader, enqueue models.Enqueue) (ok bool) {
```
pedronasser: This is ugly but we can improve it later.

ccirello: I guess we can inline that back at some point.

PR #330, opened 2016-11-22
fn: automatic release

### LINKED ISSUE #293: fn: auto release for fn
Can copy what ironcli is doing: https://github.com/iron-io/ironcli/blob/master/build-release.sh

ccirello:
Fix #293

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -28,3 +28,6 @@ docker tag $user/$service:$tag $user/$service:$version
 
 docker push $user/$service:$version
 docker push $user/$service:$tag
+
+cd fn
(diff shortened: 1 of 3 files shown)


release.sh
```diff
@@ -28,3 +28,6 @@ docker tag $user/$service:$tag $user/$service:$version
 
 docker push $user/$service:$version
 docker push $user/$service:$tag
+
+cd fn
+bash release.sh $version
```
treeder: `./release.sh` probably better?

ccirello: done

PR #332, opened 2016-11-22
functions: hot containers

ccirello:


api/runner/protocol/factory.go
```diff
@@ -0,0 +1,35 @@
...
+// Internally, a protocol must know when to alternate between stdin and stdout.
+// It returns any protocol error, if present.
+type ContainerIO interface {
+	Dispatch(stdin io.Reader, stdout io.Writer) error
+}
+
+type Protocol string
+
+const (
+	HTTP Protocol = models.FormatHTTP
```
treeder: We should still support the default, single request protocol that is in place now.

ccirello: It is. Check:
https://github.com/iron-io/functions/pull/332/files#diff-c68879dad563371d9aa7e11002087e18R99

ccirello:
@pedronasser Thanks for taking a look at this PR.

> IMHO adding more and more complexity to the IronFunctions workflow with more and more internal components dictating how the user of IronFunctions should use the platform and not letting them to extend or change those components behavior.

Indeed, I haven't added yet any extension points to the clockwork running Cold/Hot containers. I would rather wait for the actually extension need to arise than adding an extension point that become just another burden of maintenance and documentation.

However, you do have a point regarding complexity. Hot containers implementation is really more complex than cold containers. And I am open to suggestions on how to make it simpler, if you have any.

> I think a good practice is to let new components with big responsibilities to be replaceable, so the user can configure their behavior the way they want.

Care providing citation for this statement? Shouldn't components be small, sweet and to the point?

> Like I said before, I think the only two parts that must be irreplaceable are the API and the Runner. We must decide what kind of open source project we are building.

I think this is a valid debate for Slack.

pedronasser:
We can improve this in another PR.
LGTM

PR #337, opened 2016-11-23
Add appfilter and tests

pedronasser:


api/datastore/bolt/bolt.go
```diff
@@ -512,3 +514,14 @@ func (ds *BoltDatastore) Get(ctx context.Context, key []byte) ([]byte, error) {
 	})
 	return ret, nil
 }
+
+func applyAppFilter(app *models.App, filter *models.AppFilter) bool {
```
ccirello: Where is this being used?

pedronasser: Oops, forgot to use it.

PR #342, opened 2016-11-25
Add slackbot guppy example

guilhermebr:
This PR adds an example of how to run slackbots in IronFunctions.

related to issue #161


examples/slackbot/guppy/Gemfile
```diff
@@ -0,0 +1,4 @@
+source 'https://rubygems.org'
+
```
pedronasser: Missing json gem

PR #343, opened 2016-11-25
Changes on function/metric loggers

pedronasser:
Make them more plugglable

### DIFF
--- a/api/runner/async_runner.go
+++ b/api/runner/async_runner.go
@@ -51,7 +51,6 @@ func getCfg(task *models.Task) *Config {
 		Timeout: time.Duration(*task.Timeout) * time.Second,
 		ID:      task.ID,
 		AppName: task.AppName,
-		Stdout:  ioutil.Discard,
 		Env:     task.EnvVars,
 	}
 	return cfg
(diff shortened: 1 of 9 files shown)


api/runner/async_runner.go
```diff
@@ -52,8 +51,7 @@ func getCfg(task *models.Task) *Config {
 		Timeout: time.Duration(*task.Timeout) * time.Second,
 		ID:      task.ID,
 		AppName: task.AppName,
-		Stdout:  stderr,
-		Stderr:  stderr,
+		Stdout:  ioutil.Discard,
```
ccirello: Is the intention really to discard Stdout? I understand that stderr is being plugged later down the process, but stdout not.

`getCfg()` is used by `startAsyncRunners()` in order to connect both stdout and stderr. Could you please check that this behavior hasn't been changed?

pedronasser: @ccirello If read the previous impl of `getCfg()` you'll see that we was sending the output to the stderr used to log. What is the benefit of logging the async result?

ccirello: > What is the benefit of logging the async result?

So you can actually inspect what an async is doing. Or putting differently, this is a subtle change that I wouldn't mind as long as we make clear that this is the intention.

pedronasser: Oh my bad.
I'll describe it better.

ccirello: @treeder is this change OK? I know you have a more informed view on what to do with async stdout output.

PR #344, opened 2016-11-26
[WIP] Pluggable routecache

pedronasser:
The idea of this PR is to let external projects that depends on IronFunctions API, to create and plug their own cache handler.

- Routecache package not internal (in order to be used from external projects)
- Created interface called Cacher that handles the routecache. (moved methods from Server)
- All cacher methods has context.


api/server/routecache/cacher.go
```diff
@@ -0,0 +1,101 @@
...
+	if ds == nil {
+		return nil, errors.New("Missing datastore")
+	}
+	return &DefaultCacher{
+		Datastore: ds,
+		hotroutes: make(map[string]*Cache),
+	}, nil
+}
+
+func (c *DefaultCacher) PrimeCache(ctx context.Context) {
```
ccirello: IMHO, this is wrong in many levels. Going from shallow to deep.

Probably, this made in into routescache because of `cacheParetoThreshold`. Otherwise, how would the cache know how to prime itself? If that's the line of reasoning here, then it is an inversion of responsibility. Cache is meant to be used - and not to use (except, of course, for its underlying driver). Caches are not a subset/superset of databases on which you could have some sort of  "cache-level" stored procedure.

However, therein lies a deeper problem: What is this cache for? This cache was created as a way to prevent unnecessary DB conns during HTTP server `/r/` calls. This cache does hold routes - but it has nothing to do with the rest of the request service life cycle - it does not know of DB calls, it does not know of singleflights and etc.

Think about the inversion: if you accept that you can pull some sort of HTTP/Router logic _into_ the `Cacher` interface - you suddenly have two problems: a) it also means you can pull logic into other parts, such as single flight and DB drivers., how would you decide when one starts and other ends? b) how would you handle the situation on which two parts could have competing views of the same pulled logic? 

From what I see, `PrimeCache` should not exist - and should be reverted to the place it was before, operating the now injected `Cacher`.

api/server/routecache/cacher.go
```diff
@@ -0,0 +1,101 @@
...
+	"context"
+	"errors"
+	"github.com/iron-io/functions/api/models"
+	"github.com/iron-io/runner/common"
+	"math"
+	"sync"
+)
+
+type Cacher interface {
+	PrimeCache(ctx context.Context)
```
pedronasser: I'm not sure what you mean with this. I mean, with this interface the ower should do whatever they want.

api/server/routecache/cacher.go
```diff
@@ -0,0 +1,101 @@
...
+	CacheGet(ctx context.Context, appname, path string) (*models.Route, bool)
```
pedronasser: Agreed

api/server/routecache/cacher.go
```diff
@@ -0,0 +1,101 @@
...
+	RefreshCache(ctx context.Context, appname string, route *models.Route)
+	ResetCache(ctx context.Context, appname string, delta int)
+}
+
+type DefaultCacher struct {
```
pedronasser: Agreed, LRUCache would be a better name for that

api/server/routecache/cacher.go
```diff
@@ -0,0 +1,101 @@
...
+	Datastore models.Datastore
```
pedronasser: Is not only for databases, but our current implementation load routes from the database in order to build the PrimeCache.

api/server/routecache/cacher.go
```diff
@@ -0,0 +1,101 @@
...
+	hotroutes map[string]*Cache
+	mu        sync.Mutex // protects hotroutes
```
pedronasser: Sure, makes sense. Fixing it.

api/server/routecache/cacher.go
pedronasser: Makes sense.

pedronasser:
@ccirello I agree that making routecache public would add the responsibility to explain what's it works and what is its objective in the IronFunctions workflow.
In this case the routecache has the mission to prevent the runner to query for the route every request, so when the requested route is already in the hotroutes cache the datastore will not receive the query.
So when the user wants to change any kind of behavior in the datastore, he expects that everything will be requested for the datastore, but in the cases that the datastore is not consulted (eg. in this case), that behavior will not be correct. And that's the reason I'm trying to make the routecache pluggable.
IMHO we need to build IronFunctions in a form that the only things not replacable are the API and the Runner. Everything else should be injectable in order to create a platform that the users can easily modify to the use they need.

ccirello:
> In this case the routecache has the mission to prevent the runner to query for the route every request, so when the requested route is already in the hotroutes cache the datastore will not receive the query.
> So when the user wants to change any kind of behavior in the datastore, he expects that everything will be requested for the datastore, but in the cases that the datastore is not consulted (eg. in this case), that behavior will not be correct. And that's the reason I'm trying to make the routecache pluggable.

This is wrong. For every route modification, there is a cache invalidation. And if it is missing, then it is a bug that must be fixed.

WRT Cache being pluggable, you can make an extension point to Cache. And make sure that all caches, being internal or not, comply to this interface, and put all requests through this cache stack before actually checking Datastore.

The important part is to ensure that any app or route change to invalidate the cache, no matter what it is.

api/server/routecache/cacher.go
ccirello: Right - I wrote assuming a context I haven't laid down.

IMHO, the issue here is PrimeCache... It is the role of the server to prime the cache, and not that the cache should prime itself.

api/server/routecache/cacher.go
ccirello: Exactly, so probably it should be the Serve running a `primecache` function internally, that would call the plugged caches.

pedronasser:
Deferred.
This change can be done later.

Issue #339, opened 2016-11-24
Unable to run sync route execution longer than 60 seconds


denismakogon:
Container unexpectedly dies during sync execution but runs fine in docker using command after 1 minute execution:

`docker run -d -i -P denismakogon/aioservice-examples:latest`

Images: 

- [denismakogon/aioservice-examples:latest](https://hub.docker.com/r/denismakogon/aioservice-examples/tags/)
- [denismakogon/sleep120:latest](https://hub.docker.com/r/denismakogon/sleep120/)

Steps to reproduce:

`
curl -X POST -d '{"app":{"name":"testapp"}}' localhost:8080/v1/apps
`

`
curl -X POST -d '{"route":{"type": "sync", "path": "/sync", "image": "denismakogon/sleep120"}}' localhost:8080/v1/apps/testapp/routes
`

`
curl -X POST localhost:8080/r/testapp/sync
`

Function logs:
When POST to /r/{app}/route submitted
```
INFO[1442]                                               name=run.testapp.requests type=count value=1
INFO[1442]                                               name=run.testapp.waittime type=time value=0s
```

When container dies
```
INFO[1502] container status                              container_error= container_finished=2016-11-24 09:37:59.410257982 +0000 UTC container_running=false container_status=exited exit_code=137
INFO[1502]                                               name=run.testapp.error type=count value=1
INFO[1502]                                               name=run.testapp.time type=time value=1m0.06911744s
INFO[1502]                                               name="run.exec_time" type=time value=1m0.06911744s
```

But both containers are capable to run on Docker successfully.

It appears that sync func can run no longer than 60 seconds (Docker ReST client default HTTP timeout is 60 seconds).

Functions latest commit: af14904
Go version: 1.7
Docker version: 1.12.3, build 6b644ec

ccirello:
This is a bug. The problem is that there is a hard-coded 30s timeout in the source code.

https://github.com/iron-io/functions/blob/c14bc323f97f4c8d11628b27496903f80dab962e/api/server/runner.go#L172

@treeder @pedronasser @seiflotfy - what would the appropriate fix? Is the timeout an attribute of the route or the task call?

denismakogon:
@ccirello would be the right fix to set execution timeout through API? There are three cases possible:
1. No timeout specified - use default 60 seconds (or whatever it be)
2. Timeout specified - run within timeout
3. Run until not finished

cc @treeder @pedronasser @seiflotfy

ccirello:
[...]

PR #349, opened 2016-11-28
functions: fix route timeout

ccirello:
Fix https://github.com/iron-io/functions/issues/339

### DIFF
--- a/fn/routes.go
+++ b/fn/routes.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"strings"
 	"text/tabwriter"
+	"time"
 
 	functions "github.com/iron-io/functions_go"
 	"github.com/urfave/cli"
@@ -73,6 +74,11 @@ func routes() cli.Command {
 						Name:  "max-concurrency,m",
 						Usage: "maximum concurrency for hot container",
 						Value: 1,
+					},
+					cli.DurationFlag{
+						Name:  "timeout",
+						Usage: "route timeout",
+						Value: 30 * time.Second,
 					},
 				},
 			},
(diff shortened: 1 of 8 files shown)


denismakogon:
So, here the thing, we need to have persisted models definitions out of code in order to have good enough user experience during upgrades from version X to version Y.

Another question, if i would create a route before this patch and then will upgrade `functions` API up to this patch how this code will handle cases where model wouldn't have `Timeout` attribute.

Along with that, this code doesn't actually validate if `Timeout` fits `integer` range.

Mentions: https://open-iron.slack.com/archives/functions/p1480362602001126

ccirello:
> So, here the thing, we need to have persisted models definitions out of code in order to have good enough user experience during upgrades from version X to version Y.
> Another question, if i would create a route before this patch and then will upgrade functions API up to this patch how this code will handle cases where model wouldn't have Timeout attribute.

Ideally, we shall have stable model definition by v1, and from there backward compatible changes only. This has been discussed yet. We are still in the Alpha cycle where we all expect things to break often.

Currently, routes whose configuration has `Timeout` zero, will no longer have any timeout. Coding something around expected timeouts is a big mistake. Thus, it should impact anyone doing it right - and it will create a minor nuisance for those using it.

> Along with that, this code doesn't actually validate if Timeout fits integer range.

Indeed it doesn't. But also this PR is WIP. I shall align integer sizes once I regenerated client libraries.

fn/routes.go
```diff
@@ -74,6 +75,11 @@ func routes() cli.Command {
 						Usage: "maximum concurrency for hot container",
 						Value: 1,
 					},
+					cli.DurationFlag{
+						Name:  "timeout",
+						Usage: "route timeout",
+						Value: 30 * time.Second,
```
pedronasser: 60 or 30 seconds?

ccirello: done

PR #350, opened 2016-11-28
Improve listeners and handlers

pedronasser:


api/ifaces/listeners.go
```diff
@@ -7,8 +7,28 @@ import (
 )
 
 type AppListener interface {
```
ccirello: These listeners are now dealt individually. Is this being used anywhere else? If not, what do you say about removing it?

pedronasser: Removing it

PR #351, opened 2016-11-28
Add fn publish fallback option to extract appname and path

seiflotfy:


ccirello:
Open an issue with the TODO debt, please.

LGTM

PR #374, opened 2016-12-01
fn: improve how `build` handle missing version

### LINKED ISSUE #373: fn build should fail if no version in func.yaml
Either that or automatically add `version: 0.0.1`.

ccirello:
Fixes #373


ccirello:
@treeder @seiflotfy @pedronasser

PR #375, opened 2016-12-01
fn: fix max_concurrency marshal tag

ccirello:


ccirello:
@treeder @seiflotfy @pedronasser

PR #378, opened 2016-12-01
[WIP] Proposal: core concepts

pedronasser:


docs/concepts.md
```diff
@@ -0,0 +1,54 @@
...
+**Function** is the unit responsible for handling a specific part of your service. Is usually formed by a small code that receives a payload, does its work and can return the result.
+
+You can [configure a function](#Link-to-the-function-file-doc) adding a file called `func.yml` in its source directory and setup how that function needs to run.
+
+A function will run a task inside a container when triggered by a [route](#Route).
+
+### Function Structure
+
+```
+image          string                   (username/function_name)
```
ccirello: I guess this is `name` right?

docs/concepts.md
```diff
@@ -0,0 +1,54 @@
...
+
+You can easily deploy multiple functions inside the same `App` creating a `app.yml` file in the parent directory of your functions directory.
+
+### App.yml Structure
+
+```
+name: name-of-the-app 
+
+functions:
+  - "/name-of-the-app/myRoute:{path to the function directory}"
```
ccirello: Why not use regular maps in YAML?

pedronasser: Good point.

docs/concepts.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Function Structure
+
+```
+name            string                   (username/function_name)
+memory          number                   (max amount of memory used by the function)
+headers         object{key => array}     (returned HTTP headers)
+type            string (sync/async)      (how the function will be running)
+format          string (http/-)          (how the function will receive the payload)
+max_concurrency number                   (max concurrent containers can we run)
```
ccirello: What you say of making it clear that is per function per node?

docs/concepts.md
```diff
@@ -0,0 +1,55 @@
...
+
+### App.yaml Structure
+
+```
+routes:
+  /name-of-the-app/myRoute: {path to the function directory}
+  /name-of-the-app/myOtherRoute:
+    function: myOtherRoute/
+    tag:
+      private: true
```
ccirello: What is `private`?

pedronasser: Just a tag example.

docs/concepts.md
pedronasser: Sure. Updating

docs/concepts.md
ccirello: Oh... could you please then use a more example-like name? Because it felt it was some sort of IronFunctions attribute for me.

pedronasser: Oh okay. Good point

pedronasser:
The idea of this PR is define the future.

pedronasser:
Deferred

PR #379, opened 2016-12-01
fn: support for functions testing

### LINKED ISSUE #20: Make function testing framework
Probably building off this: https://github.com/iron-io/functions/issues/13

It will build the function image, then run test(s) against it.

ccirello:
Fixes #20

### DIFF
--- a/fn/glide.yaml
+++ b/fn/glide.yaml
@@ -1,5 +1,11 @@
 package: github.com/iron-io/functions/fn
 import:
+- package: github.com/aws/aws-sdk-go
+  subpackages:
+  - aws
+  - aws/credentials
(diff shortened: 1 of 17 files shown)


ccirello:
@seiflotfy @pedronasser @treeder

fn/glide.yaml
```diff
@@ -1,22 +1,22 @@
 package: github.com/iron-io/functions/fn
 import:
+- package: github.com/aws/aws-sdk-go
```
pedronasser: This dependency is used?

seiflotfy: its used in lambda.go

PR #391, opened 2016-12-05
functions: application updates no longer accept name in the body

### LINKED ISSUE #380: Unable to update app name
Create an app with name `testapp` and then try to update it.
```
curl -X PUT -d '{"app": {"name": "newapp"}}' localhost:8080/v1/apps/testapp
```
As response i get:
```
{"message":"App successfully updated","app":{"name":"testapp","config":null}}
```
And when i’m trying to list apps i get:
```
{"message":"Successfully listed applications","apps":[{"name":"testapp","config":null}]}
```
It appears that app name is immutable. So in order to fix it, following changes required:

- [ ] update docs to reflect immutability of an app name 
- [ ] at API level return HTTP 403 with error message saying that app name is immutable
- [ ] update swagger doc to reflect real use case for HTTP PUT

In swagger there's an example:
```
{
  "app": {
    "name": "string",
    "config": {}
  }
}
```
But it doesn't have any effect, would be nice to have more useful example.

ccirello:
AppUpdate was initially conceived as an upsert endpoint for apps.
It turns out that it created an inconsistency regarding updates:
updates with names divergent with URL would not actually change
application's name.

This commit atempts to address the issue by returning an HTTP
error when trying to update an application name. In swagger.yml,
application names are already `readOnly:true`. Thus there is no
change from expected behavior.

Fixes #380


ccirello:
@pedronasser @treeder @seiflotfy

denismakogon:
LGTM +1

PR #394, opened 2016-12-05
functions: fix goroutine leak in runner

### LINKED ISSUE #345: Check and fix for potential goroutine leak in api/runner
https://github.com/iron-io/functions/blob/b62a2304b02899dbcf1ad544ebeb6b8b8593cb33/api/runner/runner.go#L71

There is nothing in the main loop that breaks it in case of failure.

ccirello:
Fixes #345


ccirello:
@seiflotfy @pedronasser @treeder

PR #396, opened 2016-12-05
functions: returns HTTP error in case of route update attempt

### LINKED ISSUE #381: Unable to update route path through HTTP PUT
Similar to #380 API is not restrictive, so i can create a route
```
curl -X POST -d '{"route":{"type": "sync", "path": "/hello-sync-private", "image": "iron/hello"}}' localhost:8080/v1/apps/testapp/routes
```
And then will try to update it
```
curl -X PUT -d '{"route":{"type": "async", "path": "/hello-sync", "image": "iron/hello"}}' localhost:8080/v1/apps/testapp/routes/hello-sync-private
```
So, `type` can be modified because response shows that type is changed:
```
{"message":"Route successfully updated","route":{"app_name":"testapp","path":"/hello-sync-private","image":"iron/hello","type":"async","config":null}}
```

It appears that there's immutable attribute - `path`, but API accepts request without saying that such modification forbidden due to immutability nature of an attribute.

In order to fix issue it would be nice to have:

- [ ]  doc that reflects immutability/mutability of an attributes
- [ ] make API more restrictive by returning HTTP 403 in cases when user attempts to modify immutable attribute.

ccirello:
Ensure that attempts to update route's path are properly handled
with a HTTP error. Moreover, updates swagger file to make it
explicit that routes are immutable.

Fixes #381


denismakogon:
LGTM +1

PR #397, opened 2016-12-06
Refactor API initializer, cleaner main and more

pedronasser:
- All components that was using the `chan task.Request` also has access to the `runner` component. So I'm proposing to move that channel to the runner and the function `RunTask` (IMO they belong to the runner)
- The `server.New` was getting bigger and bigger while we add more and more injectable components to the platform, so I'm proposing adding a struct called `server.Component` and change the `server.New` signature.
- Cleaned some unecessary/unused code


api/runner/runner.go
```diff
@@ -29,6 +29,7 @@ type Runner struct {
 	flog         FuncLogger
 	availableMem int64
 	usedMem      int64
+	Tasks        chan task.Request
```
ccirello: if tasks are internal to package `runner`, why are you making it visible? Also consider that, once you make this channel something internal to package `runner`, you are also removing from test procedures to actually intercept the communication between `server` and `runner`. Is this what you want? If so, could you please elaborate what are the advantages in terms of testing of not having the possibility to intercept the communication? 

In the current test suite, we actually intercept the communication - it makes the tests faster, and prevents unnecessary interfacing with the testing docker, thus helping with separation of concerns during test execution.

api/server/runner.go
```diff
@@ -208,7 +207,7 @@ func (s *Server) serve(ctx context.Context, c *gin.Context, appName string, foun
 		c.JSON(http.StatusAccepted, map[string]string{"call_id": task.ID})
 
 	default:
-		result, err := runner.RunTask(s.tasks, ctx, cfg)
+		result, err := Api.Runner.RunTask(ctx, cfg)
```
ccirello: Another consequence of this change. We are now relying even more on something that we are trying to avoid: the `Api` global variable. https://github.com/iron-io/functions/blob/master/api/server/server.go#L23-L25

```
// Would be nice to not have this is a global, but hard to pass things around to the
// handlers in Gin without it.
```

Is this the path we want to be going to?

api/server/server.go
```diff
@@ -43,15 +43,22 @@ type Server struct {
 	Datastore    models.Datastore
 }
 
-func New(ctx context.Context, ds models.Datastore, mq models.MessageQueue, r *runner.Runner, tasks chan task.Request, enqueue models.Enqueue) *Server {
+type Components struct {
```
ccirello: I am very skeptical of this change. We are just instituting that Server has too many responsibilities: operate datastore, message queueing, container running and etc. 

And worse, this is not really handling the case in which `Components` is passed on incomplete. 

This seems to me a rather gratuitous change - that in attempting to solve an aesthetic issue, it is actually adding more issues, and more importantly, it is not solving the underlying problem of concentration of responsibilities.

I think we might be barking to the wrong tree here.

What do you think of actually making package `server` smaller? We could use concurrent programming to compose this package better, IMO. 

For the sake of exercise, I would try to refactor `server` with just a thin machinery comprising: a channel for job enqueuing and datastore.

Out of this job enqueuing job, either direct the load to a `*Runner` (also connected through a channel), or to the `MessageQueue` (perhaps also connected through a channel).

With this approach, `Server` will become significantly smaller - will have fewer dependencies, and this function signature problem will be solved as a side effect.

api/runner/runner.go
pedronasser: @ccirello I don't see why this change really prevent intercepting the communication between server and runner. Like you you've noticed the channel is still visible and can be easily replaced externally. 
If you take a look on tasks no change has been done to the test it self, we are currently testing the same way as before, but the channel creation is no longer exposed.
I see this as benefit because all packages the were using this channel also uses the `runner` and this channel is very related to the `runner`. So why not embedding that to it?
I think hiding this channel inside the runner you keep the intercept-ability but also reduce the complexity for the user when he wants to creates his own custom IronFunctions server.

api/server/runner.go
pedronasser: Yes, I agree we need a solution for that `Api` global variable. But we are actually refactoring that right now?

I mean, we could also refactor that in this PR and solve all that problem. But I think that wouldn't change the objective of this PR that is simplify and make the server instantiation simpler, mainly for users.

This PR is just fixing part of the problem, but we could also refactor everything to solve the hole problem.

api/server/server.go
pedronasser: Like I said the idea of this PR is just make it simple and clean.

We need to remember that the idea of IronFunctions is planned to be an customizable/extensible server. So if we have an API that embeds many other components (current situation, can probably get worse), the user creating its own server will have provide each other component to the API Server in its instantiation.

I agree we need to make `server` package smaller. But still the `server` will be the main component on this server and it must have access to other replaceable components.

I agree solution this adds the problem of having to assert if all components has been correctly provided in the `server.Component` (easy thing to solve IMO)

But what other solution do you have for these problems and also respect the idea of a `IronFunctions API` *extensible and customizable*?

pedronasser:
@ccirello IMHO we can't defer this to BETA1 or later.

I think this is an foundation problem of this project and needs to be solved now, so we can easily understand how we are going to build next features without disrespecting the project organization

api/runner/runner.go
ccirello: > Like you you've noticed the channel is still visible and can be easily replaced externally.

Shouldn't code be closed for modification and open for extension? Or putting differently, if you are exposing an internal part like this, you are also saying for future customizers of IronFunction that a) Tasks can be replaced at any time (which his not true); b) Tasks do not need any protection when being changed (channel itself is safe, but if you can just replace the channel with something else, then you could have a race condition here).

> If you take a look on tasks no change has been done to the test it self, we are currently testing the same way as before, but the channel creation is no longer exposed.

That's not true - or I might have read your changes incorrectly. With the external channel, there was mocks that pretended that tasks had been consumed. In this PR, you are just running these tasks in the Docker again, no?

> So why not embedding that to it?

I clearly failed to convey my concern, and I apologize. You are just stating that your change is better (embedding), but I did not see anywhere why before was worse. You essentially took a channel that was meant to link together two different components, and put in one of them - in practice defeating the meaning of it being external: show the connection of two independent executing processes.

> also reduce the complexity for the user when he wants to creates his own custom IronFunctions server.

Could you please elaborate on this? It seems something important. Perhaps some example?

api/server/runner.go
ccirello: > we are actually refactoring that right now?

Once you add _another_ use for it, then yes. If the whole refactoring did not have this line changer, I wouldn't bother. But you are adding another use for something we should be actually be trying either to get rid off or prevent.

api/server/server.go
ccirello: > if we have an API that embeds many other components (current situation, can probably get worse), the user creating its own server will have provide each other component to the API Server in its instantiation.

Aren't what abstractions for? 

> But still the server will be the main component on this server and it must have access to other replaceable components.

It seems that again I failed to make my point. If you have the design I proposed, the only components it will need is a channel of enqueueing events (bi-directional, for the sake of async) and datastore. No need to peg the runner, the mq or anything. Namely because all of these will be running separately somewhere else - serving `server` through this channel.

And, if the initialization is the problem, perhaps the solution is to create an abstraction to help anyone customizing IronFunctions.

ccirello:
@pedronasser I would like a lot to see such refactor made for alpha2. But from my perspective here, if we do it, we throw our deadline by the window. This PR will force a domino effect of other refactorings, among them making `server` smaller, adding abstractions to help customizing the `server` part of the code and whatnots.

api/runner/runner.go
pedronasser: > Shouldn't code be closed for modification and open for extension? Or putting differently, if you are exposing an internal part like this, you are also saying for future customizers of IronFunction that a) Tasks can be replaced at any time (which his not true); b) Tasks do not need any protection when being changed (channel itself is safe, but if you can just replace the channel with something else, then you could have a race condition here).

Well, the task channel was exposed either way. I still think that it could be created and kept inside the `runner` and we could easily test its functionality during runner tests.

I still don't understand why do you think keeping the task channel inside the runner prevent us from testing it correctly.

> That's not true - or I might have read your changes incorrectly. With the external channel, there were mocks that pretended that tasks had been consumed. In this PR, you are just running these tasks in the Docker again, no?

I afraid you are correct. Reviewing my changes I confirm I forgot to reproduce that behavior of mocking in most of the cases but one particular case.

But still think even with these changes, that mocking can be done with the tasks embedded in the runner.

> I clearly failed to convey my concern, and I apologize. You are just stating that your change is better (embedding), but I did not see anywhere why before was worse. You essentially took a channel that was meant to link together two different components, and put in one of them - in practice defeating the meaning of it being external: show the connection of two independent executing processes.
> Could you please elaborate on this? It seems something important. Perhaps some example?

IMO the task channel belongs to the runner and all components that were dependant to the task channel were also dependant to the `runner`. So why really keep both separated?

I agree we can improve it more by, for example, not embed the `runner` inside the `server` and just pass a channel for communication. Same for other comments that would decentralize things and keep it more simple.

Do you think we should proceed that direction?

api/server/server.go
pedronasser: What kind of abstraction you are referring?

ccirello: (Continuing from above).

One package/struct that bind all together, and use this package as an extension point for customization.

api/runner/runner.go
ccirello: I know I am being a sort of blocker. In the end, I will support you on whatever you decide. I am just trying to spark reasoning about changes.

> Same for other comments that would decentralize things and keep it more simple.

In my mind, we would benefit a lot of having smaller components that are easier to reason about, than one big component with lots of not-so-obvious internal relationships.

The downside, is that you have more dots to connect. (See comment below about that).

pedronasser:
Deffered

Issue #365, opened 2016-11-30
Documentation lagging behind on AppListeners


ccirello:
After https://github.com/iron-io/functions/pull/350, documentation is lagging behind.

https://github.com/iron-io/functions/blob/master/docs/operating/extending.md

PR #398, opened 2016-12-06
Update listener docs

pedronasser:
Closes https://github.com/iron-io/functions/issues/365

### DIFF
--- a/docs/operating/extending.md
+++ b/docs/operating/extending.md
@@ -64,7 +64,6 @@ _Triggers before and after every app updates that happens in the API_
 Triggered during requests to the following routes:
 
 - PUT /v1/apps
-- PUT /v1/apps/:app/routes
 
 ##### AppDeleteListener
 
--- a/docs/operating/extending.md
+++ b/docs/operating/extending.md
@@ -2,17 +2,61 @@
 
 IronFunctions is extensible so you can add custom functionality and extend the project without needing to modify the core.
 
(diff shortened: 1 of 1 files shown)


docs/operating/extending.md
```diff
@@ -2,17 +2,61 @@
 
 IronFunctions is extensible so you can add custom functionality and extend the project without needing to modify the core.
 
+### Replaceable Components
+
+TODO
```
ccirello: If nothing to write here, perhaps remove this section?

docs/operating/extending.md
```diff
@@ -2,17 +2,61 @@
...
+}
 ```
+
+#### Listener Requirements
+
+To be a valid listener your struct should respect interfaces combined or alone found in the file [listeners.go](/iron-io/functions/blob/master/api/ifaces/listeners.go)
+
+These are all available listeners:
+
+- AppCreateListener
```
ccirello: What do you think of adding a diagram or perhaps longe explanation for these interfaces? So people understand better exactly which part of the request cycle is being affected.

docs/operating/extending.md
pedronasser: Agreed, I'll remove it

docs/operating/extending.md
pedronasser: Okay. I'll describe each better

PR #399, opened 2016-12-06
functions: remove lru

ccirello:


ccirello:
@pedronasser @treeder

PR #404, opened 2016-12-07
Fix API inconsistencies

### LINKED ISSUE #403: Route update (HTTP PUT) modifies datastore entity by making it inconsistent
Route update (HTTP PUT) modifies datastore entity by making it inconsistent according to initial route info and requested fields to update.

Steps to reproduce:
 - create route with type `async`
 - list routes
```
ubuntu@ubuntu:~$ curl localhost:10501/v1/apps/testapp-68f72660931645198cd892/routes
{"message":"Sucessfully listed routes","routes":[{"app_name":"testapp-68f72660931645198cd892","path":"/hello","image":"iron/hello","memory":128,"type":"async","max_concurrency":1,"timeout":30,"config":{}}]}
```
 - update route
```
ubuntu@ubuntu:~$ curl -X PUT localhost:10501/v1/apps/testapp-68f72660931645198cd892/routes/hello -d '{"route":{"type":"sync"}}'
{"message":"Route successfully updated","route":{"app_name":"testapp-68f72660931645198cd892","path":"/hello","type":"sync","config":null}}
```
 - list routes
```
ubuntu@ubuntu:~$ curl localhost:10501/v1/apps/testapp-68f72660931645198cd892/routes
{"message":"Sucessfully listed routes","routes":[{"app_name":"testapp-68f72660931645198cd892","path":"/hello","image":"iron/hello","type":"sync","timeout":30,"config":{}}]}
```
 - get route info
```
ubuntu@ubuntu:~$ curl localhost:10501/v1/apps/testapp-68f72660931645198cd892/routes/hello
{"message":"Successfully loaded route","route":{"app_name":"testapp-68f72660931645198cd892","path":"/hello","image":"iron/hello","type":"sync","timeout":30,"config":{}}}
```

So there are two problems:
 - route update response doesn't match to what [swagger doc describes](https://github.com/iron-io/functions/blob/master/docs/swagger.yml#L318-L362) along with showing inconsistent info (in particular case - `config` attributes are not equal (`null` vs `{}`)
 - route update (HTTP PUT) makes route datastore entity inconsistent (`max_concurrency` is missing)

pedronasser:
The API/Datastore should now insert and update routes and apps and return consistent results.
Fixes a bug where update routes using Boltdb clear some route fields.

Fixes #403


api/models/task.go
```diff
@@ -23,7 +23,7 @@ const (
 
 const (
 	// FormatDefault ...
-	FormatDefault = ""
+	FormatDefault = "default"
```
ccirello: What happens with old routes which format is empty?

pedronasser: If DefaultFormat is empty this condition will not work correctly:
https://github.com/iron-io/functions/pull/404/files#diff-1e508e75f470efdd06b4bf58edef75a2R348

ccirello: That's not what I asked. I asked what will happen to routes already in the database, whose format is default because the field format is empty. Do you see the problem?

pedronasser: Nothing bad will happen. And if they update their route once, their format will be updated correctly.

ccirello: I am not so sure: check https://github.com/iron-io/functions/blob/master/api/runner/protocol/factory.go#L33

ccirello: It would probably fix this particular case... I propose, though, a longer check of the source code for other use cases. Perhaps, grep for the constant in the code.

pedronasser: I've searched for that constant and there's no other case.
I've just committed that change.

api/runner/protocol/factory.go
```diff
@@ -36,6 +37,8 @@ func New(p Protocol, in io.Writer, out io.Reader) (ContainerIO, error) {
 		return &HTTPProtocol{in, out}, nil
 	case Default:
 		return &DefaultProtocol{}, nil
+	case Empty:
```
ccirello: Go tip:

```Go
case Default, Empty:
```

(https://play.golang.org/p/hi1GhzW4Nv)

ccirello:
LGTM with one style suggestion.

PR #406, opened 2016-12-07
Switch to go-swagger for client lib and checks for IRON_TOKEN

treeder:
This is just for review, to see if this looks like the right way to do the auth.


pedronasser:
@treeder 
Could we replace IRON_TOKEN for just AUTH_TOKEN?
I guess some enterprise users would like to have their own authentication and might use that environment var as well.

treeder:
@pedronasser looks like I didn't answer your question, the IRON_TOKE N iss there because that's what all Iron users use right now.

treeder:
@pedronasser can you fix conflicts, then I'll review.

treeder:
@pedronasser conflicts?

pedronasser:
@treeder fixing it

pedronasser:
@treeder Waiting for this #476 to be merged

PR #407, opened 2016-12-07
[WIP] runner: check image exists locally

pedronasser:


ccirello:
@nikhilm what you say of this PR? Shouldn't this be part of iron-io/runner perhaps?

denismakogon:
Agreed with @ccirello seems like a part of runner lib. And there's obvious reason for this, all container-type specific API is encapsulated in runner to provide high-level abstraction. So, having explicit requests to Docker API would bring certain inconsistency.

api/runner/docker/client.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/docker/docker/client"
+	"golang.org/x/net/context"
+)
+
+func CheckImageLocally(image string) error {
+	cli, err := client.NewEnvClient()
+	if err != nil {
+		panic(err)
+	}
+	// Have to use ctx.Background() for now because the docker official client
```
nikhilm: Our common.Context has nothing to do with context.Context.

pedronasser:
@nikhilm 
The idea of the PR is to check if the image exists locally.
Right now we are checking only against the registry.
`driver.CanExecuteFast`, would serve for that case?

nikhilm:
You would have to write that function ;)

Best,
Nikhil




On Wed, Dec 7, 2016 at 2:19 PM -0800, "Pedro Nasser" <notifications@github.com> wrote:












@nikhilm

The idea of the PR is to check if the image exists locally.

Right now we are checking only against the registry.

driver.CanExecuteFast, would serve for that case?

pedronasser:
@nikhilm Look that PR https://github.com/iron-io/runner/pull/24

PR #408, opened 2016-12-08
Add special handler docs

pedronasser:
Closes: #408


ccirello:
I am not sure whether the phrasing could be improved as I am not an native speaker. It does LGTM however.

PR #410, opened 2016-12-08
API improvements

pedronasser:
- adds appName and routePath to the ctx in the beginning of the request.
- remove global Api object
- reduce gin dependency
- improves some error results


api/server/routes_test.go
```diff
@@ -73,7 +73,6 @@ func TestRouteDelete(t *testing.T) {
 		expectedCode  int
 		expectedError error
 	}{
-		{&datastore.Mock{}, "/v1/apps/a/routes", "", http.StatusTemporaryRedirect, nil},
```
ccirello: Why was this test removed?

api/server/runner.go
```diff
@@ -44,8 +61,6 @@ func (s *Server) handleRequest(c *gin.Context, enqueue models.Enqueue) {
 	ctx := c.MustGet("ctx").(context.Context)
 
 	reqID := uuid.NewV5(uuid.Nil, fmt.Sprintf("%s%s%d", c.Request.RemoteAddr, c.Request.URL.Path, time.Now().Unix())).String()
-	c.Set("reqID", reqID) // todo: put this in the ctx instead of gin's
```
ccirello: Isn't this used for async jobs?

api/server/server.go
```diff
@@ -19,7 +19,6 @@ import (
 
 // Would be nice to not have this is a global, but hard to pass things around to the
 // handlers in Gin without it.
```
ccirello: Please, remove the comment too.

api/server/server.go
```diff
@@ -47,13 +46,40 @@ func New(ctx context.Context, ds models.Datastore, mq models.MessageQueue, r *ru
...
+	s.Router.Use(prepareMiddleware(ctx))
+
+	return s
+}
+
+func prepareMiddleware(ctx context.Context) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		ctx, log := common.LoggerWithFields(ctx, extractFields(c))
+
+		for _, param := range c.Params {
```
ccirello: Why not use https://godoc.org/github.com/gin-gonic/gin#Params.ByName ?

docs/swagger.yml
```diff
@@ -91,6 +91,14 @@ paths:
       responses:
         200:
           description: Apps successfully deleted.
+        404:
```
ccirello: Bump version please.

api/server/runner.go
pedronasser: The async job receives it through the task object.
No other place in the code do a c.Get("reqID")

api/server/routes_test.go
pedronasser: That's not a valid route.
That falls in SpecialHandlers.

api/server/server.go
pedronasser: Because I wanted to make sure that we run those validations and store value in context only if the target route has those params (app and route).

docs/swagger.yml
pedronasser: My bad. Doing it

PR #412, opened 2016-12-09
listeners and special handlers improvements

pedronasser:
- Moved some code from `ifaces` to `server`, and removed code not being used.
- Small changes in AppListeners (all moved to app_listeners.go)
- Added RunnerListeners (listeners fired in handleRequest)
- Moved SpecialHandlers code to `special_handlers.go`
- Updated docs


pedronasser:
@ccirello Thanks for the suggestion.
I agree that the name may be wrong for some cases. 
I think it's a problem of semantics, I liked your idea of using the names `BeforeDispatch` and `AfterDispatch`.
Those names seem more qualified for their roles.

treeder:
@pedronasser are you going to update the names of those functions?

pedronasser:
@treeder done

treeder:
This broke the docs: https://github.com/iron-io/functions/blob/2a09a1c2a2ddc7e1192b316083726478d81e49a8/docs/operating/extending.md

Still has ifaces in it.

PR #417, opened 2016-12-12
Remove OpenStack alarming example in favour of Picasso

denismakogon:
Depends-On: https://github.com/iron-io/picasso/pull/27


ccirello:
@derekschultz Can we merge this?

derekschultz:
@ccirello - we can merge. One thing to note that we can come back to when the Picasso repo is public (this Fri), would be of value to link somewhere in the functions repo to the Picasso example? The reason I bring this up is because I know we had this whole idea of a /triggers directory in functions, and this PR gets rid of that, so a user browsing the functions repo will not be aware that there's an OpenStack integration, therefore let's direct them to Picasso, the OpenStack FaaS. Again, something we can come back to next week. Just raising it now. Otherwise, this is fine to merge, as the example was a little outdated anyway.

PR #428, opened 2016-12-13
Improvements on API error, swagger and status code

### LINKED ISSUE #421: Can't update a route and show better error message
I get this response:

```json
{"error":{"message":"Could not create route"}}
```

Which isn't helpful at all. 

And this is in the functions logs:

```
Dec 12 23:01:36 953f0bcf2c83 functions:  time="2016-12-13T07:01:36Z" level=error msg="Could not create route" action="server.handleRouteCreate)-fm" app=tbots error="Route already exists"  
```

Which actually has helpful information, ie: that it already exists. That said, I can't update my route to a new version because of this error.

pedronasser:
Fixes #421


api/server/routes_test.go
```diff
@@ -194,7 +194,7 @@ func TestRouteUpdate(t *testing.T) {
 		{&datastore.Mock{}, "/v1/apps/a/routes/myroute/do", ``, http.StatusBadRequest, models.ErrInvalidJSON},
 		{&datastore.Mock{}, "/v1/apps/a/routes/myroute/do", `{}`, http.StatusBadRequest, models.ErrRoutesMissingNew},
 
-		// success
+		// successTe
```
ccirello: Typo?

api/server/routes_update.go
```diff
@@ -52,7 +52,13 @@ func (s *Server) handleRouteUpdate(c *gin.Context) {
 
 	route, err := s.Datastore.UpdateRoute(ctx, wroute.Route)
 	if err != nil {
-		log.WithError(err).Debug(models.ErrRoutesUpdate)
+		if err == models.ErrRoutesNotFound {
```
ccirello: Perhaps:
```Go
if err == models.ErrRoutesNotFound {
...
} else if err != nil {
...
}
```
So to avoid if-within-if situation?

PR #431, opened 2016-12-13
plugins: add support for binary plugins

ccirello:


treeder:
This is cool, but let's wait until it's officially supported in Go and works on all platforms.

ccirello:
@treeder It is using conditional compilation for Go1.8. It is backward-compatible. So I strongly advise to have it merged anyway.

pedronasser:
I think adding plugins support (for 1.8 only) is not much a problem. And I see benefit from start working on how plugins will be used in the functions (although this should be not a priority until 1.8 out of beta).

docs/plugins.md
```diff
@@ -0,0 +1,40 @@
...
+set up an environment variable stating where the plugin are:
+
+```sh
+PLUGINS="commonlog.so" ./functions
+```
+
+Note that you can activate more than one plugin at a time using comma:
+
+```sh
+PLUGINS="commonlog.so,requestdump.so" ./functions
```
pedronasser: Is this the best solution to load plugins?
Couldn't we just add an env var pointing to a directory location from where all (valid) plugins would be loaded?
What do you think?

ccirello: It is a good idea!

seiflotfy:
Go 1.8 is not even out yet :D lets wait a bit. Also I heard that they might drop plugins from Mac :)

PR #435, opened 2016-12-14
Disable envAsHeader by default

pedronasser:
## Motive
Some enviroments (for example like mine) has some env values that causes some unexpected behaviors in the `fn call`. Users with the same kind of env might have problems running examples.

## Solution
Only call envAsHeader if the -e flag has at least one selected env variable.


ccirello:
LGTM

PR #441, opened 2016-12-14
fn: add glide support for Go build

ccirello:


treeder:
Does this run glide install for every build @ccirello ?

ccirello:
@treeder no. Only if a glide file exists.

https://github.com/iron-io/functions/pull/441/files#diff-4070a885e260010d742e6f48147e4c2dR30

PR #452, opened 2016-12-16
Go version of the client builder

pedronasser:


clients/main.go
```diff
@@ -0,0 +1,223 @@
...
+
+		// Download generated client
+		log.Printf("Downloading `%s` client...\n", language)
+		gf, err := getFile(strings.Replace(gen.Link, "https", "http", 1))
+		if err != nil {
+			log.Printf("Failed to download generated %s client. Skipping...", language)
+		}
+		ioutil.WriteFile(filepath.Join(tmpDir, "gen.zip"), gf, 0777)
+
+		// Unzip]
```
ccirello: Tiny typo.

clients/main.go
```diff
@@ -0,0 +1,223 @@
...
+		var skipFiles []string
+		tmpDir := filepath.Join(rootTmpDir, language)
+		srcDir := filepath.Join(tmpDir, "src")
+		clientDir := filepath.Join(tmpDir, fmt.Sprintf("%s-client", language))
+
+		var options map[string]interface{}
+		var deploy [][]string
+
+		// Specfic language configurations
+		switch language {
```
ccirello: Loop with a switch inside. It seems we'd be better use a map here, instead of testing for all languages against the subset that the switch block represents.

pedronasser: I don't think that's really a problem here.
The switch is there not only to check which languages is supported but also what are their specific configurations.

PR #456, opened 2016-12-22
Overwrite config/headers on update using boltdb

### LINKED ISSUE #453: Can not delete config keys
Existing config keys can not be deleted.
Affected attributes: App config, Route config, Route headers

Example.
app:
```
{
  "name":"foo", 
  "config":{
    "a":1
  }
}
```

update app with body
```
{
  "name":"foo", 
  "config":{
    "b":2
  }
}
```
will produce 
```
{
  "name":"foo", 
  "config":{
    "a":1,
    "b":2
  }
}
```
`"a"` key persists and can't be removed.

Possible solutions:
* Delete key if provided value is blank (empty string? null ?)
* Add endpoints to manage config state (more bothersome)

pedronasser:
Postgres is overwriting it right now.

Fixes #453 (To delete single keys the client will have to handle it, that's what is happening right now in `fn`)

### DIFF
--- a/api/datastore/bolt/bolt.go
+++ b/api/datastore/bolt/bolt.go
@@ -143,12 +143,7 @@ func (ds *BoltDatastore) UpdateApp(ctx context.Context, newapp *models.App) (*mo
 
 		// Update app fields
 		if newapp.Config != nil {
-			if app.Config == nil {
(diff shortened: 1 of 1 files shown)


thousandsofthem:
👍

PR #468, opened 2016-12-29
server: stats endpoint

### LINKED ISSUE #389: Stats endpoint
Include:

* Jobs running
* Jobs ran
* Jobs in sync queue

ccirello:
fixes #389


api/server/runner.go
```diff
@@ -180,6 +180,7 @@ func (s *Server) serve(ctx context.Context, c *gin.Context, appName string, foun
 		Timeout:        time.Duration(found.Timeout) * time.Second,
 	}
 
+	s.Runner.Enqueue()
```
pedronasser: I think this can generate confusion because of the method name.
I suggest using some name more descriptive like CountEnqueue() or something

PR #470, opened 2017-01-02
Cleanup main

treeder:
This removes everything unnecessary to make it easy for people to extend functions.


seiflotfy:
do we really want this as part of alpha 2?

Issue #414, opened 2016-12-11
make run-docker is buggy on linux


Gouthamve:
Running `make run-docker` twice on a linux box gives me this:

```
set -ex
docker run --rm -v /home/goutham/gocode/src/github.com/iron-io/functions:/go/src/github.com/iron-io/functions -w /go/src/github.com/iron-io/functions iron/go:dev go build -o functions-alpine
docker build -t iron/functions:latest .
Error checking context: 'no permission to read from '/home/goutham/gocode/src/github.com/iron-io/functions/data/bolt.db''.
make: *** [Makefile:14: build-docker] Error
```

I have to delete the data folder every time. Works well on the Mac though.

seiflotfy:
Will test it on Linux in a bit. Thanks for the catch :D

rdallman:
Hi @Gouthamve thanks for filing, I'm having trouble reproducing this issue (specifically as it pertains to being a bug in running in a linux environment) as outlined. It seems to be a simple permissions issue but it could be something sneaky with docker filesystem drivers or such. It seems like at a minimum whichever user you are running the command as (and maybe the `docker` user / group) needs to own the directory `/home/goutham/gocode/src/github.com/iron-io/functions` and its children, ala `$ chown -R $user:$group .`.  I think we need more info, perhaps if I post my info something will stick out as different from yours and will help you debug your system. If you could post some additional information about your directory you linked `/home/goutham/gocode/src/github.com/iron-io/functions` and your docker version we may be able to help further debug here. Here is my system info:

```
functions(master) ✗: uname -a
Linux debian 4.8.0-2-amd64 #1 SMP Debian 4.8.11-1 (2016-12-02) x86_64 GNU/Linux
functions(master) ✗: groups reed
reed : reed cdrom floppy sudo audio dip video plugdev netdev bluetooth docker
functions(master) ✗: ls -alh
total 37M
drwxr-xr-x 12 reed reed 4.0K Dec 28 18:36 .
drwxr-xr-x 20 reed reed 4.0K Dec  4 22:02 ..
drwxr-xr-x  7 reed reed 4.0K Dec 28 18:34 api
-rw-r--r--  1 reed reed  626 Dec 28 18:34 build.ps1
-rw-r--r--  1 reed reed 1.3K Dec 28 18:34 circle.yml
drwxr-xr-x  2 reed reed 4.0K Dec 28 18:34 clients
[...]

PR #471, opened 2017-01-03
fixes make run-docker

rdallman:
since the db files were being created inside of the docker container with only
permissions for the root user to rwx and docker run needs all of $PWD to be
readable in order to build a docker container on the host, `make run-docker`
was broken on any subsequent runs. If we create more permissive permissions
then we don't have that issue (group +rx)

closes https://github.com/iron-io/functions/issues/414


Makefile
```diff
@@ -34,4 +34,4 @@ run-docker: build-docker
 	set -ex
 	docker run --rm --privileged -it -e LOG_LEVEL=debug -e "DB_URL=bolt:///app/data/bolt.db" -v $(DIR)/data:/app/data -p 8080:8080 iron/functions
 
-all: dep build
\ No newline at end of file
+all: dep build
```
rdallman: intentionally leaving this in the patch ;)

PR #472, opened 2017-01-03
Add global lru for routes with keys being the appname + path

seiflotfy:
Use a global LRU for routes matching during function calls

### DIFF
--- a/THIRD_PARTY
+++ b/THIRD_PARTY
@@ -6,3 +6,7 @@ Copyright 2013 Julien Schmidt. All rights reserved. BSD-license
 For: api/server/internal/routecache/lru.go
 For: api/server/singleflight.go
 Copyright 2012 Google Inc. All rights reserved. Apache 2 license
+
+For: api/server/internal/routecache/lru.go
+For: api/server/singleflight.go
+Copyright 2012 Google Inc. All rights reserved. Apache 2 license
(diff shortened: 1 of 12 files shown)


THIRD_PARTY
```diff
@@ -5,4 +5,8 @@ Copyright 2013 Julien Schmidt. All rights reserved. BSD-license
 
 For: api/server/internal/routecache/lru.go
 For: api/server/singleflight.go
+Copyright 2012 Google Inc. All rights reserved. Apache 2 license
```
ccirello: Why the duplication here?

api/server/internal/routecache/lru.go
```diff
@@ -0,0 +1,88 @@
...
+		return
+	}
+	if ele, hit := c.cache[appname+path]; hit {
+		c.ll.MoveToFront(ele)
+		return ele.Value.(*models.Route), true
+	}
+	return
+}
+
+// Delete ...
```
ccirello: function description is missing.

api/server/routes_create.go
```diff
@@ -99,5 +99,8 @@ func (s *Server) handleRouteCreate(c *gin.Context) {
 		return
 	}
 
+	// For any create just push the new route to the front of the list
```
ccirello: Is this comment really necessary? it seems quite self-evident for me that `Refresh` will push the new route to the front of the list.

THIRD_PARTY
```diff
@@ -8,5 +8,4 @@ For: api/server/singleflight.go
 Copyright 2012 Google Inc. All rights reserved. Apache 2 license
 
 For: api/server/internal/routecache/lru.go
-For: api/server/singleflight.go
-Copyright 2012 Google Inc. All rights reserved. Apache 2 license
\ No newline at end of file
+For: api/server/singleflight.go
```
ccirello: Still duplicated...

Issue #462, opened 2016-12-29
WIP - Extending IronFunctions Spec


treeder:
We've found 3 use cases so far for extending IronFunctions:

1. API middleware, such as authentication
1. Adding extra API endpoints
1. Changing the default function route of `/r/:app/:function`, most likely for using DNS hosts in place of `/r/:app`

All of these need extra information for database operations too. 

### API Middleware

Currently, you can add middleware by accessing gin directly and adding it during setup:

```go
funcServer := server.New(ctx, ds, mqType, rnr, tasks, server.DefaultEnqueue)
funcServer.Router.Use(NewAuthMiddleware(ds))
```
We want to abstract gin out of this and provide this functionality as part of IronFunctions. 

### Adding extra API endpoints

Currently, you can add API endpoints by accessing gin and adding new handlers during setup:

```go
funcServer := server.New(ctx, ds, mqType, rnr, tasks, server.DefaultEnqueue)
funcServer.Router.GET("/something", handlerFunc)
```

We want to abstract gin out of this and provide this functionality as part of IronFunctions. 

### Changing the default function route

There is an `AddSpecialHandler(mySpecialHandler)` function available where you can add extra functions to run when a route outside of the default api routes is called. The default behavior is a 404, but you can change that with a SpecialHandler. eg:

```
funcServer.AddSpecialHandler(dnsHandler)
```

Those functions can change the internal data such as the `appName` and `path` that will be used. I don't think this is full fleshed out yet though, as we're changing these values in the context, which shouldn't be where the internal system gets it's absolutely necessary data from. 

## Proposal to handle these cases

TODO:

PR #473, opened 2017-01-04
API extension points

treeder:
This enables the user to add new endpoints to the API, that aren't part of the default API. Also simplifies main down to just a few lines as the default. 

See this for usage: https://github.com/iron-io/functions/compare/master...extensions2#diff-227391467e160aca7abf1b0671a0f807

Part of #462.

### DIFF
--- a/api/const.go
+++ b/api/const.go
@@ -4,4 +4,8 @@ package api
 const (
 	AppName string = "app_name"
 	Path    string = "path"
+
+	// Short forms for API URLs
+	CApp   string = "app"
+	CRoute string = "route"
 )
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
 	"path"
 	"sync"
 
@@ -15,7 +16,9 @@ import (
 	"github.com/ccirello/supervisor"
 	"github.com/gin-gonic/gin"
 	"github.com/iron-io/functions/api"
+	"github.com/iron-io/functions/api/datastore"
 	"github.com/iron-io/functions/api/models"
+	"github.com/iron-io/functions/api/mqs"
 	"github.com/iron-io/functions/api/runner"
 	"github.com/iron-io/functions/api/runner/task"
 	"github.com/iron-io/functions/api/server/internal/routecache"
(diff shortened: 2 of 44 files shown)


ccirello:
OK. `AppListener`. From my perspective, this is a long story. It starts as a single interface for all Listeners types, and then iteratable through all Listeners. Sometime ago, I believe in https://github.com/iron-io/functions/pull/412 , @pedronasser changed it - by my suggestion - to break into smaller interfaces, and only iterate each Listener type. This PR is reverting it.

@pedronasser any insights on this change?

treeder:
I can see the benefit of the smaller interfaces from a user perspective, but also felt more complex, more things to know about. And this is only for apps, we'll need the same thing for routes and whatever else a user will be able to do in the future.

api/const.go
```diff
@@ -4,4 +4,8 @@ package api
 const (
 	AppName string = "app_name"
```
pedronasser: Why these doesn't have C prefix?

ccirello: Why add a C prefix first off? This is not a go idiomatic standard.

treeder: @ccirello removed the C and I didn't feel like adding them back. ;)  Can figure out how we want to name the constants later.

treeder:
@ccirello @seiflotfy @pedronasser could someone review this please.

api/server/server.go
```diff
@@ -53,6 +54,24 @@ type Server struct {
 
 const cacheSize = 1024
 
+// NewEnv creates a new IronFunctions server based on env vars.
+func NewEnv(ctx context.Context) *Server {
```
pedronasser: Wouldn't be more descriptive to use the name `NewFromEnv`?

docs/operating/extending.md
```diff
@@ -15,7 +15,7 @@ package main
 
 import (
     "context"
-
+df
```
pedronasser: Should this be removed?

api/server/special_handler_test.go
```diff
@@ -1,64 +1,51 @@
...
 
 type testSpecialHandler struct{}
 
 func (h *testSpecialHandler) Handle(c HandlerContext) error {
-	c.Set(api.AppName, "test")
+	// c.Set(api.AppName, "test")
 	return nil
 }
 
 func TestSpecialHandlerSet(t *testing.T) {
```
ccirello: perhaps adding t.Skip()?

api/server/server.go
treeder: Ya, good call, changing.

seiflotfy:
Really nice work! LGTM!

PR #474, opened 2017-01-04
WIP - Middleware draft example. 

treeder:
See https://github.com/iron-io/functions/compare/extensions2...middleware?expand=1#diff-3a824a00a308fe1c8fb2b2d7fd5c2db9 for example usage.


docs/operating/extending.md
```diff
@@ -2,6 +2,13 @@
 
 IronFunctions is extensible so you can add custom functionality and extend the project without needing to modify the core.
 
+There are 4 different ways to extend the functionality of IronFunctions. 
+
+1. Listeners - listen to API events such as a route getting updated and react accordingly.
+1. Middleware - a chain of middleware is executed before an API handler is called.
+1. Add API Endpoints - extend the default IronFunctions API. 
+1. Special Handlers - TODO: DO WE NEED THIS ANYMORE??
```
pedronasser: Like I said before I think special handlers is not required.

api/server/middleware.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"net/http"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/gin-gonic/gin"
+	"github.com/iron-io/functions/api/models"
+)
+
+type Middleware interface {
+	ServeHTTP(w http.ResponseWriter, r *http.Request, app *models.App) error
```
pedronasser: How can the user use the *models.App to change the API behavior?

examples/middleware/main.go
```diff
@@ -0,0 +1,47 @@
...
+	})
+	funcServer.AddMiddleware(&CustomMiddleware{})
+
+	funcServer.Start(ctx)
+}
+
+type CustomMiddleware struct {
+}
+
+func (h *CustomMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request, app *models.App) error {
```
pedronasser: I think we should have a example that uses the *models.App to change the API behavior.

treeder: Agree with you on that, I'll have to think of an example. Closing this to create a PR against master.

Issue #467, opened 2016-12-29
Improve fn's commands organization


pedronasser:
This changes have the objective of making the fn command list clear for the new incoming user but also keep some aliases for quick management of their functions.

## Commands structure

```
  init - initialize function configuration
  
  apps - manage applications
    create - create app manually
    inspect - inspect app
    update - update app
    list - list apps
    delete - delete app
    build - create application in the API using an app.yml 
    dump - save application configuration from the API to an app.yml
    
  routes - manage routes
    call - call function remotelly
    create - create route manually
    inspect - route route
    update - update route
    list - list routes
    delete - delete route
    build - create route in the API using an functions.yml 
    dump - save route configuration from the API to an functions.yml
    
  images - manage function images
    run - run function image locally
    test - test function image
    build - build image locally using and function.yml
    bump - bump function image version
    publish - publish image to registry
    deploy - bump + build + publish + update routes
    
  lambda - manage lambda function images
    ...
```

## Aliases

```
     build    (images build)
     bump     (images bump)
     deploy   (images deploy)
     run      (images run)
     call     (routes call)
     push     (images push)
```

### Creating a new function from source
```
fn init iron/hello --runtime ruby
fn deploy myapp /hello
```

### Updating function
```
fn deploy myapp (discover route path if available in function.yml)
```

### Testing function locally
```
fn run iron/hello
```

### Testing route
```
fn call myapp /hello
```

### App management
```
fn apps create myapp
fn apps update myapp --headers "content-type=application/json"
fn apps config set log_level info
fn apps inspect myapp
fn apps delete myapp
```

### Route management
```
fn routes create myapp /hello iron/hello
fn routes update myapp /hello --timeout 30 --type async
fn routes config set myapp /hello log_level info
fn routes inspect myapp /hello
fn routes delete myapp /hello
```

PR #476, opened 2017-01-05
fn: apps commands improvements

pedronasser:
Following this issue: #467 

- Added `fn apps inspect`
- Added `fn apps update`


ccirello:
how do you do `config unset`?

pedronasser:
@ccirello right now should be like this: `fn apps routes --config myconf=,`
But that looks not so good.
How about like this? Any suggestion?
`fn apps routes --config "-myconf"`

ccirello:
If setting a configuration key to empty removes it, how do you set empty values?

treeder:
What if we do something along the lines of heroku style configs? https://devcenter.heroku.com/articles/config-vars

treeder:
config:set and config:unset

ccirello:
@treeder that's how it was before, but I am giving Pedro some room to think about this change. It was like you're suggesting, but this PR changes this.

pedronasser:
@treeder 
A command just to set and unset the config?
Couldn't it be like this?
```
fn apps update --config:set myconf --config:unset myconf2
```

pedronasser:
I added `config set` and `config unset` back.

treeder:
Are there any docs that go along with this?  Would like to see examples of these changes.

pedronasser:
@treeder Is this enough? https://github.com/iron-io/functions/issues/467

treeder:
Yes, that's great. Can you add that to the docs, that's actually a lot more clear than what's currently there: https://github.com/iron-io/functions/blob/master/fn/README.md

PR #478, opened 2017-01-06
Add routes config set/unset back

pedronasser:


fn/routes.go
```diff
@@ -419,37 +438,49 @@ func (a *routesCmd) configList(c *cli.Context) error {
...
+
+	err := a.patchRoute(appName, route, &patchRoute)
 	if err != nil {
-		return fmt.Errorf("error loading route information: %v", err)
+		return err
 	}
 
-	if msg := wrapper.Error_.Message; msg != "" {
-		return errors.New(msg)
+	fmt.Println(appName, route, "updated", key, "with", value)
```
seiflotfy: why are we doing a println? should we not be using a logger instead?

PR #490, opened 2017-01-20
Postgres Example

jmank88:
This example function provides primitive INSERT and SELECT against a postgres table through a JSON interface.  I took advantage of built in JSON capabilities to keep it general.

This was adapted from the redis example and the readme might need some more work.


examples/postgres/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Development
+
+### Building image locally
+
+```
+# SET BELOW TO YOUR DOCKER HUB USERNAME
+USERNAME=YOUR_DOCKER_HUB_USERNAME
+
+# build it
```
treeder: I know some of our examples need updating to the cli commands, but could you change the README to be like the `helloworld` examples that use the cli (much simpler).  Like this: https://github.com/iron-io/functions/tree/master/examples/hello/go

Even if just for the build, push, bump so you can get rid of the `.sh` files and simplify things a bit.

examples/postgres/test.sh
```diff
@@ -0,0 +1,43 @@
...
+sleep 5s
+
+docker run --rm -i --link test-postgres-func:postgres postgres psql -h postgres -U postgres -c 'CREATE TABLE people (first TEXT, last TEXT, age INTEGER);'
+
+RECORD1='{
+    "first": "John",
+    "last": "Smith",
+    "age": 30
+}'
+echo $RECORD1 | docker run --rm -i -e SERVER=postgres:5432 -e COMMAND=INSERT -e TABLE=people --link test-postgres-func:postgres iron/func-postgres
```
jmank88: I nearly have this converted to the simpler CLI format, but I'm having trouble converting these into 'fn run' commands.  Does 'fn run' support linking containers?  Or is there another way to expose the postgres container to the function?

jmank88: I modified my local 'fn run' to support passing through '--link' flags, although this feature is apparently deprecated, so this may not be the right approach.

I also believe I found a bug [here](https://github.com/iron-io/functions/blob/master/fn/run.go#L86), where kv[1] should be used in place of os.Getenv(). (Or perhaps a check for kv[1], with a fallback to lookup)

Are either of these changes desired?  If so, should I push the changes here, or open new PRs and keep things organized?

treeder: I like both of these changes/fixes (adding --link and fixing bug). If you make a new PR separate from this one, that would be great!

jmank88: I'll update this PR after #495 and #496 are merged.  Unfortunately, relying on 'fn init <YOUR_DOCKERHUB_USERNAME>/hello' gets in the way of using 'fn test' in place of test.sh, but there may still be a way to unravel that... (w/o passing an image override param to every command?).

treeder:
Thanks!

PR #493, opened 2017-01-23
Fix datastore error for inexistent app

### LINKED ISSUE #492: Invalid API error when access inexistent app's route
If user request for an inexistent app's route the response is `tx not writable`.
Expected error to be: `App not found` or `Route not found`

Example:
```sh
curl http://127.0.0.1:8080/v1/apps/myapp123/routes/123123
```

Response
```
ERRO[0733] Could not get route from datastore            action="server.handleRouteGet)-fm" app=myapp123 error="tx not writable" route="/123123"
```

pedronasser:
Closes #492


api/datastore/bolt/bolt.go
```diff
@@ -239,15 +239,11 @@ func (ds *BoltDatastore) GetApp(ctx context.Context, name string) (*models.App,
 }
 
 func (ds *BoltDatastore) getRouteBucketForApp(tx *bolt.Tx, appName string) (*bolt.Bucket, error) {
-	var err error
 	// todo: should this be reversed?  Make a bucket for each app that contains sub buckets for routes, etc
 	bp := tx.Bucket(ds.routesBucket)
 	b := bp.Bucket([]byte(appName))
 	if b == nil {
-		b, err = bp.CreateBucket([]byte(appName))
```
treeder: Can you verify we're not expecting this behavior anywhere else?  I think you're right that we shouldn't create an app when getting a route, but want to make sure anything using this method is ready for the change.

PR #496, opened 2017-01-24
make fn run pass through --link flags

jmank88:
Passing through '--link' flags from 'fn run' to 'docker run' as discussed in #490.


fn/testfn.go
```diff
@@ -134,7 +134,7 @@ func runlocaltest(target string, in, expectedOut, expectedErr *string, env map[s
 		restrictedEnv = append(restrictedEnv, k)
 	}
 
-	if err := runff(target, stdin, &stdout, &stderr, restrictedEnv); err != nil {
+	if err := runff(target, stdin, &stdout, &stderr, restrictedEnv, nil); err != nil {
```
jmank88: Note that I didn't add test support, though it might be pretty trivial from what I saw. I'm not sure if that alone would be sufficient to allow example/postgres/test.sh to be a 'fn test' configuration instead (since those exist in the func.yaml, which isn't checked in), but there may be other compelling reasons.

ccirello:
Thanks for this PR. I do see a problem with it, however. You are adding a new parameter (`--link`) for running functions locally - but it lacks the implementation on the Docker driver. It is important that `fn run` and `fn call` both be feature parity. 

Also, I believe in the long term, we aspire for an agnostic approach to the containers. I wonder how this would reflect on other possible drivers down the road, such rkt. 

Any ideas on this @seiflotfy @pedronasser @treeder ?

treeder:
@ccirello `link` on call would probably have to be different since the link would have to be made while starting the `iron/functions` container. So really `link` is only valid on `fn run` for testing. With regards to rocket and others, I don't really see a way around not having the link to do local testing so may have to have it regardless of future tech we support.

ccirello:
@treeder makes sense.

pedronasser:
Agreed with @treeder

treeder:
Thanks @jmank88 !

PR #502, opened 2017-01-30
Middleware

treeder:
Rebasing on master instead of extensions2, continues #474 

You can see example usage here: https://github.com/iron-io/functions/pull/502/files#diff-3a824a00a308fe1c8fb2b2d7fd5c2db9R21

### DIFF
--- a/api/server/middleware.go
+++ b/api/server/middleware.go
@@ -3,6 +3,7 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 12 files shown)


api/server/middleware.go
```diff
@@ -0,0 +1,107 @@
...
+	context.Context
+
+	ginContext  *gin.Context
+	nextCalled  bool
+	index       int
+	middlewares []Middleware
+}
+
+func (c *middlewareContextImpl) Next() {
+	fmt.Println("NEXT called")
```
ccirello: Is this necessary? Or perhaps could be replace with logrus?

api/server/middleware.go
```diff
@@ -0,0 +1,107 @@
...
+	c.nextCalled = true
+	c.index++
+	c.serveNext()
+}
+
+func (c *middlewareContextImpl) serveNext() {
+	fmt.Println("serveNext()", c.Index())
```
ccirello: Ditto.

docs/operating/extending.md
```diff
@@ -2,6 +2,13 @@
 
 IronFunctions is extensible so you can add custom functionality and extend the project without needing to modify the core.
 
+There are 4 different ways to extend the functionality of IronFunctions. 
+
+1. Listeners - listen to API events such as a route getting updated and react accordingly.
+1. Middleware - a chain of middleware is executed before an API handler is called.
+1. Add API Endpoints - extend the default IronFunctions API. 
+1. Special Handlers - TODO: DO WE NEED THIS ANYMORE??
```
ccirello: No. I don't think we need it.

treeder: I'll remove special handlers in another PR.

api/server/middleware.go
treeder: Cleaning up, thanks.

PR #510, opened 2017-02-03
Removed special handlers, added runMiddleware.

treeder:


docs/operating/extending.md
```diff
@@ -70,38 +70,18 @@ See examples of this in [/examples/extensions/main.go](/examples/extensions/main
...
-## Special Handlers
-
-To understand how **Special Handlers** works you need to understand what are **Special Routes**.
-
-**Special Routes** are routes that doesn't match any other API route. 
-
-With **Special Handlers** you can change the behavior of special routes in order to define which function is going to be executed.
-
-For example, let's use special handlers to define `mydomain` as the `appname` for any request for `mydomain.com`.
+There are two separate sets of middleware, one for the IronFunctions API and one that wraps requests to user created endpoints.
```
Gouthamve: A little misleading? As only the API endpoints are wrapped by middleware not the functions invokers: https://github.com/iron-io/functions/blob/ce26f665ea2b1e83ce093ba0f9fe86d27b055e11/api/server/server.go#L252-L271

Or is it supposed to encapsulate the functions endpoints too?

api/server/server.go
```diff
@@ -42,9 +43,9 @@ type Server struct {
 
 	apiURL string
 
-	specialHandlers []SpecialHandler
 	appListeners    []AppListener
 	middlewares     []Middleware
+	runMiddlewares  []Middleware
```
Gouthamve: Can Run Middlewares take the `task.Config` as a parameter? I mean if we are middlewaring the runner, we would want to change the config right?

Or is there way to map the context to the task config?

api/server/runner.go
```diff
@@ -29,42 +28,37 @@ type runnerResponse struct {
...
-		log.WithError(err).Errorln("Error using special handler!")
-		c.JSON(http.StatusInternalServerError, simpleError(errors.New("Failed to run function")))
-		return
+	fctx := &middlewareContextImpl{
+		Context: ctx,
 	}
+	fctx.app = app
+	// fctx.index = -1
+	fctx.ginContext = c
+	fctx.middlewares = s.runMiddlewares
```
Gouthamve: Okay. I was thinking this was a middleware for the runner. Why are we using it on the special handler?

Did I just get the whole intention wrong?

treeder:
Closing, #532 replaces this.

Issue #512, opened 2017-02-04
Docker and curl commands in functions/README.md don't work.


jconning:
~~`curl get.iron.io/fn` needs a -L so it can follow the redirect.~~
`${pwd}` in docker run should actually be `${PWD}`.

Fix the README.md so that these commands work out of the box.

treeder:
The fn part of this is a dupe of #501 .

jconning:
These are both valid curl commands that do different things:
`curl -sSL`
`curl --ssl`

Maybe I'm the only nutcase who would ever be confused by this but what the heck I changed it to be `curl -LSs`.

Issue #501, opened 2017-01-30
QuickStart guide instructions for downloading fn don't work.


jconning:
The Quickstart guide in functions/README.md says to install fn with "curl -sSL http://get.iron.io/fn | sh".  This doesn't work on my MacBook Pro running OS X Yosemite 10.10.5 because curl doesn't follow redirects by default.  But add the -L option and it works: "curl -L http://get.iron.io/fun | sh".

treeder:
The README has the -L option already in `-sSL` so it should have already worked if that was the issue. 

`curl -sSL http://get.iron.io/fn | sh`

derks:
Tested on latest macOS Sierra, and the original command `curl -sSL ...` **does** work just fine.  Not sure if that is an issue with cURL on Yosemite specifically... maybe providing output of the failing command might help?

jconning:
I got confused and thought that `-sSL` was supposed to be `--ssl`.  Seriously I saw "sSL" and immediately thought SSL, secure sockets layer.  It didn't occur to me that it was actually `-s -S -L`.

False bug report!  But my eyes would be more pleased if it were `-s -S -L` or `-LSs`.

@treeder @derks

PR #513, opened 2017-02-04
Doc fixes: change docker and curl commands so they will execute.

jconning:
Fixes issue #512.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ configuration options [here](docs/options.md). If you are on Windows, check [her
 Install the IronFunctions CLI tool:
 
 ```sh
-curl -L http://get.iron.io/fn | sh
+curl -LSs http://get.iron.io/fn | sh
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,9 +89,11 @@ configuration options [here](docs/options.md). If you are on Windows, check [her
 Install the IronFunctions CLI tool:
 
 ```sh
-curl -sSL http://get.iron.io/fn | sh
+curl -L http://get.iron.io/fn | sh
```
treeder: See #501, the `-sSL` already has `-L` in it, I'm not sure why this didn't work for you. cc @rdallman

jconning: My bad.  I saw "-sSL" and immediately thought secure sockets layer, so I typed "--ssl".

jconning:
I changed the curl flags to `curl -LSs`

PR #523, opened 2017-02-14
Dynamic routes

jmank88:
This PR is a work in progress towards adding dynamic routes #77.
I validated my understanding of the problem and the solution I sketched out on #77 by first updating the mock Datastore.  Then I wrote more tests and some benchmarks.  The postgres implementation was a little more complicated than I first thought, but it's working (check my tests for any missed edge cases!).  I'll start on Bolt next.

## TODO:

[x] Mock
[x] Tests
[x] Benchmarks
[x] Postgres
[x] Bolt
[x] Cleanup

## Benchmarks

I wrote a handful of benchmarks to measure route lookup speed in a few situations:
- standard: varying numbers of apps and 'standard' routes
- segmentCount: varying numbers of path segments - /test /test/test 
- branchFactor: vary numbers of paths per segment/node - /test1 /test2/ /test3
- segmentSize: varying numbers of characters per path segment - /a /aa /aaa

### Mock

This one is pretty snappy as expected.  It will be interesting to see how it maps onto Bolt.

```
BenchmarkDatastore/GetRoute/standard/apps=100/routes=10-4         	 3000000	       569 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=100-4        	 3000000	       578 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=1000-4       	 2000000	       705 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=10000-4      	 2000000	       724 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=10-4          	 3000000	       486 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=100-4         	 3000000	       585 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=1000-4        	 2000000	       642 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=10000-4       	 2000000	       679 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=10-4           	 3000000	       466 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=100-4          	 3000000	       535 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=1000-4         	 3000000	       556 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=10000-4        	 2000000	       633 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=1-4             	10000000	       190 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=5-4             	 3000000	       428 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=10-4            	 2000000	       698 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=50-4            	  500000	      2902 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=100-4           	  200000	      5689 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=1-4             	10000000	       197 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=5-4             	 5000000	       219 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=10-4            	10000000	       219 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=50-4            	10000000	       258 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=100-4           	10000000	       218 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=500-4           	10000000	       227 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=1000-4          	 5000000	       229 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=5000-4          	10000000	       235 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=10000-4         	10000000	       239 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=1-4                 	 5000000	       447 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=5-4                 	 3000000	       433 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=10-4                	 3000000	       503 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=50-4                	 1000000	      1085 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=100-4               	 1000000	      1674 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=200-4               	  500000	      2954 ns/op

```

### Postgres

I haven't investigated the limiting factors here, but it's a starting point, and it would be interesting to see the difference a real postgres instance makes, compared to the on-demand local container used here.

In general, these numbers would be greatly improved through the use of a custom index, but this regex approach is much simpler and more portable.

```
BenchmarkDatastore/GetRoute/standard/apps=100/routes=10-4         	    5000	    255551 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=100-4        	    1000	   2069528 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=1000-4       	    1000	   1929290 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=10000-4      	     500	   2244895 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=100000-4              300       4623863 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=10-4          	    5000	    378152 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=100-4         	    1000	   2074248 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=1000-4        	    1000	   2190993 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=10000-4       	     500	   4118233 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=10-4           	    3000	    432084 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=100-4          	     500	   2427358 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=1000-4         	     500	   3858346 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=10000-4        	     100	  14040583 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=1-4             	    5000	    297277 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=5-4             	    3000	    335625 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=10-4            	    3000	    384372 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=50-4            	    2000	    987397 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=100-4           	    1000	   1977562 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=1-4             	    5000	    346381 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=5-4             	    5000	    329776 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=10-4            	    3000	    346472 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=50-4            	    2000	    699068 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=100-4           	    2000	    767996 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=500-4           	    1000	   1089479 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=1000-4          	    1000	   1642663 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=5000-4          	     300	   5156957 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=10000-4         	     200	   9478381 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=1-4                 	    5000	    287341 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=5-4                 	   10000	    234003 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=10-4                	    5000	    243681 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=50-4                	    3000	    423030 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=100-4               	    2000	    697696 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=200-4               	    1000	   2105683 ns/op
```

### Bolt

```
BenchmarkDatastore/GetRoute/standard/apps=100/routes=10-4         	  100000	     10394 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=100-4        	  100000	     11161 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=1000-4       	  100000	     11792 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=10000-4      	  100000	     12266 ns/op
BenchmarkDatastore/GetRoute/standard/apps=100/routes=100000-4                 3000        620729 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=10-4          	  100000	     10154 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=100-4         	  100000	     10718 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=1000-4        	  100000	     11782 ns/op
BenchmarkDatastore/GetRoute/standard/apps=10/routes=10000-4       	    3000	    501588 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=10-4           	  100000	     10026 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=100-4          	  100000	     10925 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=1000-4         	  100000	     11803 ns/op
BenchmarkDatastore/GetRoute/standard/apps=1/routes=10000-4        	    3000	    499176 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=1-4             	  200000	      8758 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=5-4             	  200000	     10325 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=10-4            	  100000	     13882 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=50-4            	   30000	     42406 ns/op
BenchmarkDatastore/GetRoute/segmentCount/segments=100-4           	   20000	     75791 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=1-4             	  200000	      7748 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=5-4             	  200000	      7778 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=10-4            	  200000	      7797 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=50-4            	  200000	      9140 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=100-4           	  200000	      8018 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=500-4           	  200000	      8145 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=1000-4          	  200000	      8279 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=5000-4          	  200000	      8536 ns/op
BenchmarkDatastore/GetRoute/branchFactor/branches=10000-4         	  100000	     10752 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=1-4                 	  200000	     10066 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=5-4                 	  100000	     10360 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=10-4                	  100000	     10975 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=50-4                	  100000	     13087 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=100-4               	  100000	     16011 ns/op
BenchmarkDatastore/GetRoute/segmentSize/chars=200-4               	  100000	     21656 ns/op
```

**UPDATE** Throwing up some Bolt numbers now that it's functional: It appears to be roughly 10x slower than the mock, and roughly 100x faster than postgres.
After I do some cleanup I'll re-run them all fresh and update.
**UPDATE 2** New numbers for all three.
**UPDATE 3** 100,000 routes for bolt and postgres.


treeder:
Great stuff!  Keep in mind we'll probably want to support other databases in the future too so techniques that apply to any database preferred. I suspect what you do with Bolt may solve that problem though.

api/models/datastore.go
```diff
@@ -5,21 +5,66 @@ import (
 	"errors"
 )
 
+//TODO path name validation: implementations currently assume/require/rely on:
+// - leading slash
+// - overall legal/scrubbed path (no relative paths, consecutive slashes, etc)
+// - trailing slash allowed and distinct from none
```
jmank88: Is there existing documentation on what exactly is a valid path?

api/models/datastore.go
```diff
@@ -5,21 +5,66 @@ import (
...
+	// InsertApp inserts an App. Returns ErrDatastoreEmptyApp when app is nil, and
+	// ErrDatastoreEmptyAppName when app.Name is empty.
+	// Returns ErrAppsAlreadyExists if an App by the same name already exists.
 	InsertApp(ctx context.Context, app *App) (*App, error)
+
+	// UpdateApp updates an App's Config. Returns ErrDatastoreEmptyApp when app is nil, and
+	// ErrDatastoreEmptyAppName when app.Name is empty.
+	// Returns ErrAppsNotFound if an App is not found.
+	//TODO are app.Config's fields added to the existing field or do they replace them?
 	UpdateApp(ctx context.Context, app *App) (*App, error)
```
jmank88: Looking at this again, it seems like ```Config``` should be completely replaced. Otherwise, in order to delete fields, some sort of empty/null/nil-value-means-delete would have to be devised.

Whatever is decided here also likely applies to ```UpdateRoute``` as well.

api/models/datastore.go
```diff
@@ -5,21 +5,66 @@ import (
...
+
+	// RemoveApp removes the App named appName. Returns ErrDatastoreEmptyAppName if appName is empty.
+	// Returns ErrAppsNotFound if an App is not found.
+	//TODO remove routes automatically? #528
```
jmank88: This seems straightforward. Any reason not to include with this PR? #528

api/models/datastore.go
```diff
@@ -5,21 +5,66 @@ import (
...
 	RemoveApp(ctx context.Context, appName string) error
 
+	// GetRoute looks up a matching Route for appName and the literal request route routePath.
+	// Returns ErrDatastoreEmptyAppName when appName is empty, and ErrDatastoreEmptyRoutePath when
+	// routePath is empty.
+	// Returns ErrRoutesNotFound when no matching route is found.
+	//TODO should these return ErrAppsNotFound or RoutesNotFound? (or the unused 'ErrRoutesGet')
```
jmank88: Any preference?

api/models/datastore.go
```diff
@@ -5,21 +5,66 @@ import (
...
+	// InsertRoute inserts a route. Returns ErrDatastoreEmptyRoute when route is nil, and ErrDatastoreEmptyAppName
+	// or ErrDatastoreEmptyRoutePath for empty AppName or Path.
+	// Returns ErrRoutesAlreadyExists if the exact route.Path already exists, or ErrRoutesCreate if a conflicting
+	// route already exists.
 	InsertRoute(ctx context.Context, route *Route) (*Route, error)
+
+	// UpdateRoute updates route's Config and Header fields. Returns ErrDatastoreEmptyRoute when route is nil, and
+	// ErrDatastoreEmptyAppName or ErrDatastoreEmptyRoutePath for empty AppName or Path.
+	//TODO are route's .Config and .Header fields merged with or do they replace existing fields?
 	UpdateRoute(ctx context.Context, route *Route) (*Route, error)
```
jmank88: Similar question to ```UpdateApp```.  Additionally, do empty string route fields indicate 'set field empty', or 'do not update this field`.

api/models/route.go
```diff
@@ -125,6 +125,51 @@ func (r *Route) Validate() error {
...
+	if new.Config != nil {
+		if r.Config == nil {
+			r.Config = make(Config)
+		}
+		for k, v := range new.Config {
+			r.Config[k] = v
+		}
+	}
+}
+//TODO are these sql LIKE queries? or strict matches?
```
jmank88: I'm guessing ```LIKE``` query since that's much more useful.

api/server/routes_get.go
```diff
@@ -14,7 +14,10 @@ func (s *Server) handleRouteGet(c *gin.Context) {
 
 	appName := c.MustGet(api.AppName).(string)
 	routePath := path.Clean(c.MustGet(api.Path).(string))
-
+	//TODO does this validation make sense? move to validator?
+	// -removes trailing slash
+	// -permits non-rooted paths
+	// -permits '.', '..', ('../')*
```
jmank88: I'm thinking that ```datastoreutil.validator``` should have centralized/shared path validation, and maybe it makes sense to include cleaning there as well.  Part of validation would be to reject an un-cleaned path anyways.

jmank88:
@treeder @pedronasser  I posted some comments that I could use input on before cleaning up the last few open items.

api/models/datastore.go
treeder: I believe you can patch configs, so it shouldn't be a complete replace. Eg: `fn apps update --config DB_URL=http://example.org/ myapp` . That will update DB_URL only and leave the rest as is. 

Regarding the rest of these things, why not return nil if the app doesn't exist like it used to?  An app that doesn't exist isn't really an error. 

Also, What's the difference between a nil app and a ErrAppsNotFound?

jmank88: Patching should be doable. (The current postgres impl for this and UpdateRoute both currently act as 'setters', but shouldn't become too complicated to fix).

The distinction between nil and not found is regarding the ```app``` parameter vs. ```App``` the type - a nil parameter is an illegal argument, while trying to update an App by name that doesn't exist will result in that particular App not being found.  Maybe it's confusing for the doc to reference the param name as ```app``` would ``` `app` ``` make it more clear?

When you say ```like it used to```, which implementation are you referring too?  Most of these error docs are just documenting existing behavior, or trying to guess at which is correct between inconsistent implementations or method behavior (for example, the postgres impl already returned ErrAppsNotFound when no rows were updated).  Without a broader perspective on how this API fits into the project (or even if it's public?) I don't have strong opinions either way, and these TODOs are sort of the points where I couldn't align the impls without deciding one way or the other, so whatever you think is best.

jmank88: The CI build brought some more tests to my attention that might resolve some of these.

jmank88: Patching ```map[string]string``` types is straightforward, since empty string values can signal deletes.  However, ```Route.Header``` is ```map[string][]string```, so assuming we patch the values by concatenating the arrays (or set union), what would be the protocol for deleting keys or individual value array elements?

I suppose an empty array value could signal 'delete key', but it would be unfortunate to have to nuke a key and reinsert all but one array element in order to remove a single element.

Am I overlooking something simple?

api/models/datastore.go
seiflotfy: Yeah maybe we should define more strict rules since I was able to create routes such as "///" and "/?"

seiflotfy: I think this can be dealt with in another issue...

api/server/routes_get.go
seiflotfy: +1 on the validator. And again this can be done in a different PR

api/models/datastore.go
seiflotfy: GetRoute and GetRoutes should return the same thing and have the same logic and be verbose about it. If the "App" is missing then return `ErrAppsNotFound` else return `RoutesNotFound`

api/models/datastore.go
seiflotfy: So I think returning an ErrAppsNotFound is better IMHO when trying to update the config of an app that is not found. For now I think its good to idea to just send in an empty array value for deletion... Nice one

api/models/datastore.go
jmank88: This can be a separate PR. The service layer prohibits it anyways.

api/models/route.go
jmank88: Separate PR?

api/models/datastore.go
jmank88: No problem waiting for another issue, though both inserting and querying these sorts of routes could have very strange behavior in the mean time.  Especially characters significant to regexp. (those regexps may need some additional escaping depending on what is allowed)

jmank88:
@seiflotfy @treeder This is in a fairly stable state now. Would you prefer issues for these ```//TODO```s be written up now? Or are the comments record enough?

seiflotfy:
@jmank88 can you open new issues with the TODO?

jmank88:
@seiflotfy I believe everything is accounted for between those two and #528

treeder:
@jmank88 seeing as how this is a very large PR and hard to fully grok, could you provide more insight as to how you solved this in the generic case that doesn't have features like regex queries?

jmank88:
I had an insight this morning that could greatly simplify and speed up the bolt implementation.  I originally dismissed the idea of using a single flat bucket with some sort of composite keys, but now I think that with the right path encoding we can use prefix scans to crawl the route tree. TBD - but something like url-encoded ASCII, with '/' replaced by a byte that sorts first or last, and other tricks if ':' and '*' are allowed as 'regular' path chars.  If this works out, then the abstract tree.go stuff may no longer be necessary, and both bolt and mock would be much smaller and straightforward.

jmank88:
No actionable update, but for posterity I wanted to note that I resolved the merge conflicts and the diff is much cleaner/smaller now (down from ```+2600/-1400``` to ```+1800/-600```) and there is still more to cherry pick out.  The current test failures are expected, since the new Redis datastore does not support the dynamic routes used in the common tests.

api/server/routes_update.go
```diff
@@ -39,6 +39,12 @@ func (s *Server) handleRouteUpdate(c *gin.Context) {
 	wroute.Route.AppName = c.MustGet(api.AppName).(string)
 	wroute.Route.Path = path.Clean(c.MustGet(api.Path).(string))
 
+	if err := wroute.Validate(true); err != nil {
```
seiflotfy: can't this logic be moved into a new PR?

jmank88: Yeah.  I'll pick out all the server stuff today.

jmank88: #566

ccirello:
It seems @jmank88 is taking the gradual approach to this PR. Closing this as it is on a conflicting state. This can be revived later in other PRs.

PR #531, opened 2017-02-17
Updates to enable local testing without Docker registry

### LINKED ISSUE #348: How to use without Dockerhub
Can't seem to locate the docs, even the fn tool seems to imply that you need a docker hub username. Seems like you should be able to:

1) Use local images without pulling registry. Kubernetes pods for example have options for specifying imagePullPolicy: Always | IfNotPresent | Never.

2) Use a custom registry

But can't seem to find the docs about enabling either of these.

### LINKED ISSUE #514: How to add a route to an app with a local image?
I want to add a route to an app with a local image which is not pushed to hub.docker.com or is pushed to a private registry like "registry.private.com". I can't find a way to configure the registry path and it always add "registry.hub.docker.com" in front of the image name.

treeder:
Removes image check when updating routes to make local testing possible without having to push to docker hub. Also mounts the docker socket in the quickstart to enable this. 

Thoughts?

Closes #514 
Closes #348

### DIFF
--- a/api/server/routes_create.go
+++ b/api/server/routes_create.go
@@ -7,7 +7,6 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/iron-io/functions/api"
 	"github.com/iron-io/functions/api/models"
-	"github.com/iron-io/functions/api/runner/task"
 	"github.com/iron-io/runner/common"
 )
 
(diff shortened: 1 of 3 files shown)


Gouthamve:
Instead of removing the check entirely, can we check the local images and then query the registry images and fail if the image doesn't exist in both places?

treeder:
The check I removed on creating and updating the routes should probably be removed regardless, it makes that API call too slow. I'm thinking it should just be done at run time now (which I believe is already the case).

api/server/routes_create.go
```diff
@@ -44,13 +43,13 @@ func (s *Server) handleRouteCreate(c *gin.Context) {
...
-		c.JSON(http.StatusBadRequest, simpleError(models.ErrUsableImage))
-		return
-	}
+	// err = s.Runner.EnsureImageExists(ctx, &task.Config{
+	// 	Image: wroute.Route.Image,
+	// })
+	// if err != nil {
+	// 	c.JSON(http.StatusBadRequest, simpleError(models.ErrUsableImage))
+	// 	return
+	// }
```
seiflotfy: Can we turn this into a flag --check-image-exists instead

treeder: Ya, we could. I'm not sure if it would be used much though, it really slows down the call, seconds instead of milliseconds.

seiflotfy: Then lets just remove the blocks completely ?

treeder: Good call. Testing what happens when we don't check and an image exists...

treeder: Going to make a new issue to follow up on this. Currently if the image isn't found, the response to the user is a blank 200. Created #540 to follow up on this.

treeder:
Can I get an approval or what?

PR #532, opened 2017-02-18
Middleware and requests improvements

pedronasser:
- Removed the necessity of two contexts, now using just *gin.Context (that respects context.Context)
- MiddlewareContext is basically a wrapper for the *gin.Context that controls the flow of the middleware execution.
- Added the RequestController. It's an interface that controls and hold the request object, response writer, requested app name, requested route path, and the request logger.

@treeder

### DIFF
--- a/api/server/request_controller.go
+++ b/api/server/request_controller.go
@@ -1,27 +1,36 @@
 package server
 
 import (
+	"context"
 	"net/http"
+	"sync"
 
-	"github.com/Sirupsen/logrus"
 	"github.com/gin-gonic/gin"
 	"github.com/iron-io/functions/api"
+	"github.com/iron-io/functions/api/models"
 	"github.com/iron-io/runner/common"
 )
 
 type RequestController interface {
-	AppName() string
-	SetAppName(string)
-	RoutePath() string
-	SetRoutePath(string)
+	App() *models.App
+	SetApp(models.App)
+	Route() *models.Route
+	SetRoute(models.Route)
 	Request() *http.Request
 	Response() http.ResponseWriter
-	Logger() logrus.FieldLogger
-	SetLogger(logrus.FieldLogger)
+	Error() error
 }
 
 type requestInfo struct {
-	ctx *gin.Context
+	ctx   *gin.Context
+	app   *models.App
+	route *models.Route
+	err   error
+
+	appOnce       sync.Once
+	lazyLoadApp   func() (*models.App, error)
+	routeOnce     sync.Once
+	lazyLoadRoute func() (*models.Route, error)
 }
 
 func (r *requestInfo) Response() http.ResponseWriter {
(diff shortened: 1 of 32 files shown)


api/server/request_controller.go
```diff
@@ -0,0 +1,117 @@
...
+
+type RequestController interface {
+	AppName() string
+	SetAppName(string)
+	RoutePath() string
+	SetRoutePath(string)
+	Request() *http.Request
+	Response() http.ResponseWriter
+	Logger() logrus.FieldLogger
+	SetLogger(logrus.FieldLogger)
```
denismakogon: Should middleware be able to override logger?

api/server/request_controller.go
```diff
@@ -0,0 +1,117 @@
...
+		return l
+	}
+	return nil
+}
+
+func (s *Server) wrapHandler(f func(*gin.Context, RequestController)) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		log := common.Logger(c)
+
+		c.Set(api.AppName, c.Param(api.CApp))
```
treeder: Please don't add these types of things to the context, I removed all of them to avoid inappropriate use of the context. This is what the RequestController is for isn't it?

api/server/routes_create.go
```diff
@@ -89,9 +86,9 @@ func (s *Server) handleRouteCreate(c *gin.Context) {
 
 	}
 
-	route, err := s.Datastore.InsertRoute(ctx, wroute.Route)
+	route, err := s.Datastore.InsertRoute(c, wroute.Route)
 	if err != nil {
-		handleErrorResponse(c, err)
+		handleErrorResponse(c, r, err)
```
treeder: Do internal methods need the RequestController?  Can't they just use the gin context?

api/server/routes_update.go
```diff
@@ -1,20 +1,17 @@
...
 	"github.com/iron-io/functions/api/models"
 	"github.com/iron-io/functions/api/runner/task"
-	"github.com/iron-io/runner/common"
 )
 
-func (s *Server) handleRouteUpdate(c *gin.Context) {
-	ctx := c.MustGet("ctx").(context.Context)
-	log := common.Logger(ctx)
+func (s *Server) handleRouteUpdate(c *gin.Context, r RequestController) {
+	log := r.Logger()
```
treeder: The context has the logger, why get it from the RequestController?

api/server/server.go
```diff
@@ -258,33 +223,32 @@ func (s *Server) bindHandlers(ctx context.Context) {
 	engine.GET("/stats", s.handleStats)
 
 	v1 := engine.Group("/v1")
-	v1.Use(s.middlewareWrapperFunc(ctx))
```
treeder: Why did you add wrapHandlers instead of just using the one middleware?

api/server/request_controller.go
treeder: I think @denismakogon is right, can we remove both of the logger methods?  Use context for the logger like it was.

api/server/request_controller.go
```diff
@@ -0,0 +1,117 @@
...
+	"net/http"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/gin-gonic/gin"
+	"github.com/iron-io/functions/api"
+	"github.com/iron-io/runner/common"
+)
+
+type RequestController interface {
+	AppName() string
```
treeder: Why can't we just have the full App here?

api/server/request_controller.go
```diff
@@ -0,0 +1,117 @@
...
+	SetAppName(string)
+	RoutePath() string
```
treeder: Same for Route?

treeder:
I meant to choose "request changes"

api/server/request_controller.go
pedronasser: @treeder @denismakogon 
Since logger is kind a important request-unique value, I thought would be a good practice to make it visible to the middleware instead of making the user to find the logger from the the context.Value (and receive an interface{})

api/server/request_controller.go
pedronasser: Why really have a full app here? I mean, the full App struct has just name a config. The config is really necessary for this situation? 
This objects holds request required information and the only two information I think right now is required for the request is appName and routePath.
Make sense?

api/server/request_controller.go
pedronasser: Same response from above

api/server/request_controller.go
pedronasser: @treeder since right now we are using *gin.Context as the main request context and if you check *gin.Context implementation (https://github.com/gin-gonic/gin/blob/master/context.go), you will se that it stores any context value using `Set()` and get using `Get()`.
So why not use it?
When I set the `api.AppName` value here, any middleware will be able to get that value using `ctx.Value(api.AppName).(string)`

api/server/routes_create.go
pedronasser: Yeah, we can use just *gin.Context, but is it that bad to use it? And I think it provides clearly what kind of information an API request handler can have.
Should we remove it?

api/server/routes_update.go
pedronasser: Same response from above

api/server/server.go
pedronasser: Oh, I think we can revert it to middleware.
I was using wrapper because I using it to the background context to the request, but since we are just using *gin.Context, is not needed anymore.
I'll revert it

api/server/request_controller.go
treeder: What if they need more information about the App?  Or want to set config/env vars, etc.

pedronasser: @treeder If you follow the request flow there's no App (unless its name) and Route (unless its path)  information until the request reaches the final request handler. So from where would, that information, come from? As far as I know, there's no database before the final handler, maybe inside middlewares. Are you thinking about middlewares doing those database calls, and writing that App/Route struct so in the final handler, instead of calling the database again, it would just use the struct?

Sorry I'm not understanding the need of a App/Route struct in this situation

Gouthamve: I agree with @treeder about passing down the populated App initially itself.

Usecase: suppose I have a middleware that based on the caller (identified from a HEADER) wants to change the limits (Memory/Timeout) of the container. This cannot be via the middleware now.

denismakogon: @treeder @Gouthamve 
Another use cases - if have a header that represents a user ID (X-Project-ID), and i want to restrict user from accessing other apps by passing list of apps (that are stored in DB, in mapping user_id ---> app) to include/exclude. So that'll be very useful with App(s) objects (at least i suppose so).

treeder: @pedronasser my pull requests filled them in properly, copy from there.

api/server/request_controller.go
treeder: Because, that's not what context is for (I really hope we can put that discussion to rest at some point). And we're passing the App itself along now (the entire reason for doing this was to remove things from the context) so why would they ever use this in the context?

api/server/routes_create.go
treeder: sure, leave it for now.

api/server/request_controller.go
treeder: I think we remove these, the logger is optional in the middleware and I don't think we want to lock ourselves into these functions on the interface forever (unless we're 100% sure, which we're not).

PR #534, opened 2017-02-18
Config EnvVars passed without leading _

Gouthamve:
* Earlier X=Y was passed as _X=Y


treeder:
thanks!

seiflotfy:
There was a reason we had the _ ? Trying to find out why again

jmank88:
Is some prefix necessary to prevent clashes with vars like ```METHOD``` and ```ROUTE```?

seiflotfy:
LGTM can you check for any documentation that needs changing.

Issue #540, opened 2017-02-21
If image doesn't exist, no error is returned to user


treeder:
If image is not found in the registry at run time, the user gets a blank 200 response. It should return an error, 404 or something.

seiflotfy:
the merged code changed the return to 401
```ERRO[0003] Credentials not authorized, trying next.      error="Get https://registry.hub.docker.com/v2/iron/notexist/manifests/latest: http: non-successful response (status=401 body=\"{"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":[{"Type":"repository","Class":"","Name":"iron/notexist","Action":"pull"}]}]}")" image="iron/notexist" server= username=```

Gouthamve:
Umm, I think the error should be 500. Because the docker client is probably not authorised and not the client making the request.

seiflotfy:
I think it should be 404 since the image does NOT EXIST

Gouthamve:
Umm, the route exists (its mapped to an image and is registered). But the server internally is not able to run the image and errors out.

404 is for routes that are not even registered. But again, this is just minor and I guess we can go with 404 also for now.

seiflotfy:
```fn routes create myapp /notexist iron/notexist``` the image does not exist
the path is not listed under `myapp`

seiflotfy:
@Gouthamve I agree it should return 500, the path is registered so 404 is not ideal

seiflotfy:
@Gouthamve WDYT?

Gouthamve:
Yep! This sounds good!

treeder:
Hmmm, that sounds right, 5XX of some sort is probably most appropriate.

PR #546, opened 2017-02-27
Return 500 if trying to run path with non-existing image

seiflotfy:
Fixes: #540

### DIFF
--- a/api/server/runner.go
+++ b/api/server/runner.go
@@ -231,6 +231,7 @@ func (s *Server) serve(ctx context.Context, c *gin.Context, appName string, foun
 					Message: err.Error(),
 				},
 			})
+			log.WithError(err).Error("Failed to run task")
 			break
 		}
 		for k, v := range found.Headers {
--- a/api/server/runner.go
+++ b/api/server/runner.go
@@ -225,6 +225,12 @@ func (s *Server) serve(ctx context.Context, c *gin.Context, appName string, foun
 	default:
 		result, err := runner.RunTask(s.tasks, ctx, cfg)
 		if err != nil {
+			c.JSON(http.StatusInternalServerError, runnerResponse{
+				RequestID: cfg.ID,
+				Error: &models.ErrorBody{
+					Message: err.Error(),
+				},
+			})
 			break
 		}
 		for k, v := range found.Headers {


api/server/runner.go
```diff
@@ -225,6 +225,12 @@ func (s *Server) serve(ctx context.Context, c *gin.Context, appName string, foun
 	default:
 		result, err := runner.RunTask(s.tasks, ctx, cfg)
 		if err != nil {
```
treeder: We need to log this so the error isn't lost and someone can figure out what's wrong.

seiflotfy: its already in the logs from the runner AFAICT
```ERRO[0003] Failed to pull image                          action="server.handleRunnerRequest)-fm" app=myapp call_id=357e93d0-af46-5cbb-83c6-7f8628fb2344 error="API error (404): {\"message\":\"repository iron/notexist not found: does not exist or no pull access\"}\n" image="iron/notexist" registry="https://registry.hub.docker.com" route="/lol2" username=```

treeder:
Oh, ok. I'm thinking the runner probably shouldn't be logging that, it's up to the caller to decide what to do with the error. And what if at some point in the future, the runner decides to stop logging that?  Then it's lost in the void and turns into a debugging nightmare. cc @rdallman 

Recent Twitter thread I noticed on this: https://twitter.com/peterbourgon/status/834350104043741184

rdallman:
@treeder we log and return the same error, wrt text we slightly improve the error before returning it so they're not the same value, but it's the same contextually. I don't really find twitter posts from some random guy as principles (yes i know who he is, we're all random guys). there are multiple cases in the runner where we want to log the error and it may also end up getting returned, due to the nature of retrying a lot of the docker calls, where we really do want to log it at the source. that being said, libraries that don't log stuff are ideal, but sometimes it's impractical. it would probably be nice to let callers specify some kind of logger output object in case they don't want logs, but at the same time this driver imo will never be used outside of our own devices. tl;dr doesn't really seem like a big deal ?

treeder:
No, not a big deal at all, just saying we should log it too, just in case runner decides to not log it in the future and the error disappears into the void at that point.

api/server/runner.go
seiflotfy: Done!

PR #547, opened 2017-02-28
Docker login 

### LINKED ISSUE #506: Support private images on Docker Hub
Something along the lines of:

```
fn docker login ...
```

See (or copy) https://github.com/iron-io/ironcli

Runner should already support this, we'll have to pass the info into it though.

BupycHuk:
closes #506


fn/login.go
```diff
@@ -0,0 +1,89 @@
...
+		Context: context.Background(),
+	}
+
+	res, err := l.client.Docker.PostDockerLogin(params)
+	fmt.Fprintf(os.Stdout, "res := %v", res)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "err: %v", err)
+		return fmt.Errorf("unexpected error: %v", err)
+	}
+	fmt.Println(`Added docker repo credentials`)
```
treeder: Why backticks?

treeder (changes requested):
I wonder about the security implications of this. It may be that this features requires an authentication module or something?

BupycHuk:
@treeder 
I'm going to encrypt docker credentials using AES or RSA.
Do you think this is enough?

api/server/server.go
```diff
@@ -69,11 +71,16 @@ func NewFromEnv(ctx context.Context) *Server {
 
 	apiURL := viper.GetString(EnvAPIURL)
 
-	return New(ctx, ds, mq, apiURL)
+	auth := auth.DockerAuth{
+		Datastore: ds,
+		Key:       []byte("AES256Key-32Characters1234567890"),
```
treeder: User should pass this in through env var perhaps?

treeder:
Maybe we start here with global docker credentials, then once we add some authentication middleware, those could also have per user credentials.

treeder:
@BupycHuk has some conflicts here.

BupycHuk:
@treeder i have fixed conflicts and add new tests, could you take a look, pls

PR #548, opened 2017-02-28
Add support for redis as a datastore

seiflotfy:
Fixes: #388


api/datastore/redis/redis.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	hset := fmt.Sprintf("routes:%s", route.AppName)
+	if reply, err := ds.conn.Do("HGET", hset, route.Path); err != nil {
+		return nil, err
+	} else if reply == nil {
+		return nil, models.ErrRoutesNotFound
+	}
+
+	return ds.setRoute(hset, route)
```
jmank88: Should this be patching ```reply``` with ```route```'s fields?

api/datastore/redis/redis.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	hset := fmt.Sprintf("routes:%s", appName)
+	if n, err := ds.conn.Do("HDEL", hset, routePath); err != nil {
+		return err
+	} else if n == 0 {
+		return models.ErrRoutesRemoving
+	}
+
+	if _, err := ds.conn.Do("SREM", "routes", hset); err != nil {
```
jmank88: Does this call remove the whole set of routes for the app?

seiflotfy: since Redis is a dictionary based DB you need to keep track of all the existing "sets" so you can iterate through them later... check ```GetRoutes```

jmank88: Maybe I'm misunderstanding. Isn't ```routes``` a set of names of per app route sets, in the form of ```routes:<appName>```? If so, doesn't this operation make the route set for this app inaccessible when iterating over ```routes```?  In other words, doesn't it leave an inconsistent state unless we happen to be removing the last route from this app?

seiflotfy: no its just an index of what apps I have, I could have gotten it via HKEYS on "apps".... maybe you got a point

jmank88: Right, that's what I meant. The app is dropped from the index, even if this isn't the last route for that app.

seiflotfy: Got rid of them

itamarhaber:
@seiflotfy :heart:

seiflotfy:
@jmank88 you lovely person ❤️

Issue #418, opened 2016-12-13
Support MySQL as DB and RabbitMQ as MQ persistance for smoother OpenStack integration


denismakogon:
In order to get tighter integration with OpenStack infrastructure it would be nice to have support of:

- [ ] MySQL as DB storage layer
- [ ] RabbitMQ as MQ storage layer

PR #551, opened 2017-03-01
Datastore tests

jmank88:
This PR cherry picks some datastore tests and bug fixes from #523, since it's on hold and these pieces are useful now (#388 , #418).  Additionally, it narrows the size and scope of #523, whatever it ends up becoming.

This is currently stable. There are more things worth pulling from 523 (server updates, validator, etc.), but they can be spread across smaller PRs if desired.


api/datastore/postgres/postgres.go
```diff
@@ -562,3 +606,17 @@ func (ds *PostgresDatastore) Get(ctx context.Context, key []byte) ([]byte, error
...
+	if err != nil {
+		return err
+	}
+	err = f(tx)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+	return tx.Commit()
+}
```
jmank88: Does anyone have experience with the proper way to handle these transactions?

@seiflotfy and I are wondering if we need to account for the situation where ```f(tx)``` panics (L616).  If the process will terminate anyways, then we may not need to worry about an explicit ```Rollback()``` here.  On the other hand, if we expect something to recover higher up in the stack, then we should probably defer recover and take care of it.

jmank88: Here is an alternative:
```go
func (ds *PostgresDatastore) Tx(f func(*sql.Tx) error) (err error) {
	var tx *sql.Tx
	if tx, err = ds.db.Begin(); err != nil {
		return
	}
	defer func() {
		if r := recover(); r != nil {
			_ = tx.Rollback()
		} else if err != nil {
			_ = tx.Rollback()
		} else {
			err = tx.Commit()
		}
	}()
	err = f(tx)
	return
}
```

seiflotfy: Yeah I like that one more....  I think @treeder would have some good input there. Else I really like it 👍

treeder:
Thanks @jmank88 !

PR #552, opened 2017-03-01
redis test ping loop

jmank88:
Replaces static wait with a ping loop for redis docker container.


api/datastore/redis_test.go
```diff
@@ -16,8 +18,26 @@ func prepareRedisTest() func() {
...
+	for {
+		c, err := redis.DialURL(tmpRedis)
+		if err == nil {
+			_, err = c.Do("PING")
+			if err == nil {
+				c.Close()
+				break
+			}
+		}
+		c.Close()
```
jmank88: This may not be necessary.

Issue #415, opened 2016-12-11
Provide capability to retrieve task by specific call ID


denismakogon:
Since async route execution returns `call_id` from user perspective it would be useful to get execution task instead of querying all tasks by sending HTTP GET to `/tasks`

Proposal
---------

Implement new HTTP route `/tasks/{call_id}`

Returns:
[Single task entry data](https://github.com/iron-io/functions/blob/master/docs/swagger.yml#L447-L494)

HTTP errors:

- **HTTP 404** Task not found

Aside change:

Consider to bring consistency in naming: async execution return `call_id`, but in order to get info users are querying for `/tasks`. So, it is necessary to change `call_id` to `task_id` for naming consistency.

ccirello:
I am sorry, the usefulness it not evident for me. Could you please put forward a concrete case in which this is a superior replacement for the current flow?

denismakogon:
@ccirello current there's no way to get info on specific task, in order to accomplish task retrieving it is required to iterate over all existing tasks, for some cases it would take some time to get proper one. Isn't it a valid reason to have new route?

ccirello:
@denismakogon what's the intent here? I am afraid I do not follow your reasoning. The worker alone should be the entity talking to `/tasks` endpoint. And it must be able to read the task it's been assigned to. Why do you need to have this entrypoint? Which component will be talking to it, and do you expect this component to achieve with the payload that, in theory, only interests to the worker?

denismakogon:
@ccirello The idea was to provide capability to get exit status of a function to users, since task payload includes exit status, we can get it and send back to users as part of response to status check for async function execution by its `call_id` and then i've faced with problem - i can get all tasks (assume that IronFuntions is shared between multiple users in OpenStack, in my particular case - 6 users trying to run their functions). So, that's why i created this issue - for simplifying process of getting task status. Am i missing something or using `/tasks` in wrong way?

denismakogon:
@treeder @seiflotfy @pedronasser @derekschultz guys, may i ask you for an input on this topic? We still don't have a mechanism for IronFn users to track their async executions and may be increase time to keep task data for a specific execution a bit longer?

denismakogon:
Related to https://github.com/iron-io/functions/issues/528

treeder:
I think this makes sense, but currently we don't keep track of any task status, other than outputting it to the logs. We probably do need to have this for async tasks, otherwise there's no way to check if a task was successful or not (other than parsing logs). 

[...]

PR #553, opened 2017-03-01
Switch from call_id to task_id for consistency

denismakogon:
As part of async execution HTTP API response IronFunctions returns `call_id`. But it's actually referring to task ID (from `models.Task`). In order to keep consistency between tasks objects (that are available via `/tasks` HTTP API) that are containing `task_id` and no `call_id` it is necessary to rename `call_id` to `task_id`.

This patch renames `call_id` that comes as part of API response on async execution request to `task_id` for API consistency purposes.

HTTP API impact: `call_id` renamed to `task_id`

Partially addresses: #415


treeder:
Is this only for async functions?  We named them call_id in reference to a "functions calls", not sure we should have task_id and call_id.

denismakogon:
@treeder the thing is that Async execution returns call_id, but I need to query `/tasks` to get info on it. So, for implementing `/tasks/{task_id}` we need to clarify what we return to a users, would it be call_id or taks_id that can be tracked via `/tasks`

denismakogon:
@seiflotfy @pedronasser @treeder Guys, can we make certain decision on this one?

seiflotfy:
So I am pretty ok with that change, we are not 1.0 and I like the changes to the docs.
I'd like to have a +1 from @pedronasser @treeder @ccirello  
Then lets push it through. Also can you make sure you add a note in the changelogs?

seiflotfy (approved):
LGTM

ccirello:
A good merge message covering on the reason and consequences of this change is a must. Regarding everything else, LGTM.

treeder:
The main thing that bugs me about this is having different names for essentially the same thing. We should choose one or the other across the board.

denismakogon:
@treeder isn't `task_id` fits better here rather than `call_id`?

ucirello (changes requested):
I am just trying to decouple my account from this.

PR #556, opened 2017-03-01
Use SSL install url for fn cli

adelevie:
I know the other URL is shorter, but this is much safer.


README.md
```diff
@@ -89,7 +90,7 @@ configuration options [here](docs/options.md). If you are on Windows, check [her
 Install the IronFunctions CLI tool:
 
 ```sh
-curl -LSs http://get.iron.io/fn | sh
+curl -LSs https://goo.gl/VZrL8t | sh
```
treeder: @adelevie what do you think of this instead?

Issue #451, opened 2016-12-16
Setup Chocolatey for fn tool installs on Windows


treeder:
https://chocolatey.org/

Looks like a homebrew for windows.

martinpinto:
Hi @treeder, I've created an installer package using the documentation from [chocolatey](https://chocolatey.org/docs/create-packages) and taken into consideration all rules. I've tested it and it successfully installed it on my windows machine. In order for me to push this to chocolatey, I have to create a key and push it from the console. But before I do, I wanted to check with you guys. I will gladly post a link to the repo for review purposes if required.

treeder:
@martinpinto nice!  Could you make a pull request here?  Maybe put the files under /fn/chocolatey directory.

treeder:
Or share the repo you have.

treeder:
We'll probably have to do the initial upload though so we can own chocolatey package name if you don't mind?

martinpinto:
I created a new PR #564 as requested. In order to create the package and test it a `choco pack functions.nuspec` command is required.

PR #564, opened 2017-03-02
#451 Added chocolatey package support for Functions.

martinpinto:
Added a new package for chocolatey for the latest Functions version.


fn/chocolatey/functions.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <id>functions</id>
+    <version>0.2.25</version>
+    <packageSourceUrl>https://github.com/iron-io/functions/</packageSourceUrl>
+    
+    <title>Iron.io Functions (Install)</title>
+    <authors>Iron.io Inc.</authors>
+    <projectUrl>https://github.com/iron-io/functions/</projectUrl>
+    <!--<iconUrl>http://cdn.rawgit.com/__REPLACE_YOUR_REPO__/master/icons/functions.png</iconUrl>-->
+    <copyright>2017 Functions Iron.io</copyright>
+    <tags>serverless faas docker lambda</tags>
```
denismakogon: IFAIK we'd like to keep ourselves for being associated with Serverless (see https://serverless.com/). So it'll be better to have next tags:
`functions faas docker containers` (as an alternative - include COE there).

fn/chocolatey/functions.nuspec
```diff
@@ -0,0 +1,21 @@
...
+<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>functions</id>
+    <version>0.2.25</version>
+    <packageSourceUrl>https://github.com/iron-io/functions/</packageSourceUrl>
+    
+    <title>Iron.io Functions (Install)</title>
+    <authors>Iron.io Inc.</authors>
+    <projectUrl>https://github.com/iron-io/functions/</projectUrl>
```
denismakogon: trailing slash? just remove slash right after `functions`

fn/chocolatey/functions.nuspec
treeder: I'd say keep serverless and add functions. We do want to be associated with the term serverless as that's what everyone is referring to it as for the time being.

fn/chocolatey/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,24 @@
+﻿$ErrorActionPreference = 'Stop';
+
+$packageName= 'functions'
+$toolsDir   = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$url        = 'https://github.com/iron-io/functions/releases/download/0.2.25/fn.exe'
```
treeder: Could this URL always grab the latest release?  Similar to `install.sh` ?

fn/chocolatey/functions.nuspec
martinpinto: ok done. Removed both the trailing slash from the `<projectUrl>` and `<packageSourceUrl>`.

fn/chocolatey/functions.nuspec
martinpinto: ok done. Kept `serverless` and added `functions`.

fn/chocolatey/tools/chocolateyinstall.ps1
martinpinto: Unfortunately this URL is bound to that version. We might need to adjust scripts to auto increase the version. Updated version to 0.2.36.

denismakogon (approved):

fn/chocolatey/tools/chocolateyinstall.ps1
treeder: Yes, we need to make the scripts update the version on release, same way as we do for git, docker, cli install, etc. 

How does this get submitted to Chocolatey?  We'll want to make sure that's automated too.

treeder (changes requested):

PR #565, opened 2017-03-02
Datastore validator

jmank88:
This change cherry picks the common datastore validator from #523 to DRY up parameter validation from all the ```models.Datastore``` implementations.  No test behavior was changed, only mock initialization.

Notice that all implementations now have the same or better validation than before, and although there were inconsistencies between them, they all pass the same tests in both cases.  This means there are some untested edge cases possibly worth addressing.

### DIFF
--- a/api/datastore/bolt/bolt.go
+++ b/api/datastore/bolt/bolt.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Sirupsen/logrus"
 	"github.com/boltdb/bolt"
 	"github.com/iron-io/functions/api/models"
+	"github.com/iron-io/functions/api/datastore/internal/datastoreutil"
 )
 
 type BoltDatastore struct {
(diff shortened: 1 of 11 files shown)


seiflotfy:
Can you make sure that empty val should trigger delete when possible,
rather than storing an empty value. Then we are good

jmank88:
> Can you make sure that empty val should trigger delete when possible,
rather than storing an empty value. Then we are good

So the only related usage is:

> Put(ctx, []byte("test"), nil)

Which then confirms that ```Get``` returns a nil/empty value. So getting something that doesn't exist isn't an error, it just returns nil/empty anyways. So doing a delete behind the scenes would just be for efficiency, and wouldn't be visible through the API.

treeder:
CircleCI isn't updating these statuses very well lately.

treeder:
Thanks again @jmank88 !  These are awesome contributions.

PR #566, opened 2017-03-03
server updates

jmank88:
This change picks out more pieces of #523.

- Splits route validation and defaults into separate methods to leverage from route create and update.
- Improved and additional route test cases.

### DIFF
--- a/api/models/route.go
+++ b/api/models/route.go
@@ -87,6 +87,7 @@ func (r *Route) SetDefaults() {
 	}
 }
 
+// Validate validates field values, skipping zeroed fields if skipZero is true.
 func (r *Route) Validate(skipZero bool) error {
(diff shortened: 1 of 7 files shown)


api/models/route.go
```diff
@@ -114,7 +84,56 @@ func (r *Route) Validate() error {
...
+		}
+	}
+
+	if !skipZero || r.Format != "" {
+		if r.Format != FormatDefault && r.Format != FormatHTTP {
+			res = append(res, ErrRoutesValidationInvalidFormat)
+		}
+	}
+
+	//TODO negative concurrency?
```
jmank88: Any reason not to add this now?

seiflotfy: while you are at it! Why not?

PR #573, opened 2017-03-07
Test harness to assess whether fnlb works properly

jconning:
This is a test harness whose purpose is to generate requests against fnlb so that we can assess whether fnlb is correctly distributing requests across multiple nodes.  It expects multiple IronFunctions nodes to be running and one fnlb instance.  It uses Docker's HOSTNAME environment variable as a container id -- the function reads HOSTNAME to determine what container it is running in and the test harness tracks how many times each node was called.

In the current implementation, a single route is called many times.  I envision subsequent versions that create many routes to verify that given routes are routed to the correct instances.

Question for reviewers:
- The test harness depends on a specially instrumented function (jconning/primes:0.0.1).  What is the best practice to get that added to DockerHub as iron/primes?


treeder:
@jconning this is awesome!  Will review shortly.

test/fnlb-test-harness/primes-func/func.go
```diff
@@ -0,0 +1,58 @@
...
+					numLoops, _ = strconv.Atoi(value)
+				}
+			}
+		}
+	}
+
+	for i := 0; i < numLoops; i++ {
+		primes := sieveOfEratosthenes(maxPrime)
+		_ = primes
+		if i == numLoops - 1 {
```
seiflotfy: just get rid of this :D it does nothing

test/fnlb-test-harness/primes-func/func.go
```diff
@@ -0,0 +1,58 @@
...
+					maxPrime, _ = strconv.Atoi(value)
+				}
+				if key == "loops" {
+					numLoops, _ = strconv.Atoi(value)
+				}
+			}
+		}
+	}
+
+	for i := 0; i < numLoops; i++ {
```
seiflotfy: I am not sure what this part is for honestly nothing is changing in this for loop

test/fnlb-test-harness/main.go
```diff
@@ -0,0 +1,128 @@
...
+		fmt.Printf("  %s %d\n", node, len(ex))
+	}
+}
+
+func discoverContainerIds() {
+	// Discover the Docker hostname of each node; create a mapping of hostnames to host/port.
+	// This is needed because IronFunctions doesn't make the host/port available to the function (as of Mar 2017).
+	fmt.Println("Discovering container ids for every node (use Docker's HOSTNAME env var as a container id)...")
+	for _, s := range nodes {
+		e, err := executeFunction(s, route, 100, 1)
```
seiflotfy: just nitpicking but a nice way of writing this is
```go
if e, err := executeFunction(s, route, 100, 1); err == nil {
} else {
}
```

jconning:
I'll be addressing the code review comments soon!

test/fnlb-test-harness/primes-func/func.go
jconning: The loop simply repeats the calculation of the prime numbers. Nothing changes, it just calculates them again. The reason for doing this is so we can increase the work for the CPU without increasing the memory usage.

The function enables the amount of work performed by the CPU to increase so we can assess whether longer running functions have any impact on fnlb. The function provides two ways to control how much work is done by the CPU: you can set the maximum prime number (maxPrime) and you can set the number of times that the entire calculation will be repeated (numLoops).  Increasing the maximum prime number causes the memory usage to increase, while increasing the number of repetitions does not.

test/fnlb-test-harness/primes-func/func.go
jconning: See above.

jconning:
@seiflotfy I have made the changes you requested.

PR #575, opened 2017-03-08
Fix #418 Added MySQL as DB storage layer.

### LINKED ISSUE #418: Support MySQL as DB and RabbitMQ as MQ persistance for smoother OpenStack integration

martinpinto:
Added MySQL driver to IronFunctions

### DIFF
--- a/api/datastore/mysql/mysql.go
+++ b/api/datastore/mysql/mysql.go
@@ -49,14 +49,14 @@ type rowQuerier interface {
 }
 
 /*
-MySQLDatastore ...
+MySQLDatastore Defines a basic MySQL Datastore struct.
 */
 type MySQLDatastore struct {
 	db *sql.DB
 }
 
 /*
-New ...
+New Creates a new MySQL Datastore.
 */
 func New(url *url.URL) (models.Datastore, error) {
 	u := fmt.Sprintf("%s@%s%s", url.User.String(), url.Host, url.Path)
@@ -89,7 +89,7 @@ func New(url *url.URL) (models.Datastore, error) {
 }
 
 /*
-InsertApp ...
+InsertApp Inserts an app to MySQL.
 */
 func (ds *MySQLDatastore) InsertApp(ctx context.Context, app *models.App) (*models.App, error) {
 	var cbyte []byte
@@ -117,7 +117,7 @@ func (ds *MySQLDatastore) InsertApp(ctx context.Context, app *models.App) (*mode
 }
 
 /*
-UpdateApp ...
+UpdateApp Updates an existing app on MySQL.
 */
 func (ds *MySQLDatastore) UpdateApp(ctx context.Context, newapp *models.App) (*models.App, error) {
 	app := &models.App{Name: newapp.Name}
@@ -174,7 +174,7 @@ func (ds *MySQLDatastore) UpdateApp(ctx context.Context, newapp *models.App) (*m
 }
 
 /*
-RemoveApp ...
+RemoveApp Removes an existing app on MySQL.
 */
 func (ds *MySQLDatastore) RemoveApp(ctx context.Context, appName string) error {
 	_, err := ds.db.Exec(`
@@ -190,7 +190,7 @@ func (ds *MySQLDatastore) RemoveApp(ctx context.Context, appName string) error {
 }
 
 /*
-GetApp ...
+GetApp Retrieves an app from MySQL.
 */
 func (ds *MySQLDatastore) GetApp(ctx context.Context, name string) (*models.App, error) {
 	row := ds.db.QueryRow(`SELECT name, config FROM apps WHERE name=?`, name)
@@ -229,7 +229,7 @@ func scanApp(scanner rowScanner, app *models.App) error {
 }
 
 /*
-GetApps ...
+GetApps Retrieves an array of apps according to a specific filter.
 */
 func (ds *MySQLDatastore) GetApps(ctx context.Context, filter *models.AppFilter) ([]*models.App, error) {
 	res := []*models.App{}
@@ -257,7 +257,7 @@ func (ds *MySQLDatastore) GetApps(ctx context.Context, filter *models.AppFilter)
 }
 
 /*
-InsertRoute ...
+InsertRoute Inserts an route to MySQL.
 */
 func (ds *MySQLDatastore) InsertRoute(ctx context.Context, route *models.Route) (*models.Route, error) {
 	hbyte, err := json.Marshal(route.Headers)
@@ -322,7 +322,7 @@ func (ds *MySQLDatastore) InsertRoute(ctx context.Context, route *models.Route) 
 }
 
 /*
-UpdateRoute ...
+UpdateRoute Updates an existing route on MySQL.
 */
 func (ds *MySQLDatastore) UpdateRoute(ctx context.Context, newroute *models.Route) (*models.Route, error) {
 	var route models.Route
@@ -391,7 +391,7 @@ func (ds *MySQLDatastore) UpdateRoute(ctx context.Context, newroute *models.Rout
 }
 
 /*
-RemoveRoute ...
+RemoveRoute Removes an existing route on MySQL.
 */
 func (ds *MySQLDatastore) RemoveRoute(ctx context.Context, appName, routePath string) error {
 	res, err := ds.db.Exec(`
@@ -443,7 +443,7 @@ func scanRoute(scanner rowScanner, route *models.Route) error {
 }
 
 /*
-GetRoute ...
+GetRoute Retrieves a route from MySQL.
 */
 func (ds *MySQLDatastore) GetRoute(ctx context.Context, appName, routePath string) (*models.Route, error) {
 	var route models.Route
@@ -460,7 +460,7 @@ func (ds *MySQLDatastore) GetRoute(ctx context.Context, appName, routePath strin
 }
 
 /*
-GetRoutes ...
+GetRoutes Retrieves an array of routes according to a specific filter.
 */
 func (ds *MySQLDatastore) GetRoutes(ctx context.Context, filter *models.RouteFilter) ([]*models.Route, error) {
 	res := []*models.Route{}
@@ -487,7 +487,7 @@ func (ds *MySQLDatastore) GetRoutes(ctx context.Context, filter *models.RouteFil
 }
 
 /*
-GetRoutesByApp ...
+GetRoutesByApp Retrieves a route with a specific app name.
 */
 func (ds *MySQLDatastore) GetRoutesByApp(ctx context.Context, appName string, filter *models.RouteFilter) ([]*models.Route, error) {
 	res := []*models.Route{}
@@ -561,7 +561,7 @@ func buildFilterRouteQuery(filter *models.RouteFilter) (string, []interface{}) {
 }
 
 /*
-Put ...
+Put Inserts an extra into MySQL.
 */
 func (ds *MySQLDatastore) Put(ctx context.Context, key, value []byte) error {
 	_, err := ds.db.Exec(`
@@ -582,7 +582,7 @@ func (ds *MySQLDatastore) Put(ctx context.Context, key, value []byte) error {
 }
 
 /*
-Get ...
+Get Retrieves the value of a specific extra from MySQL.
 */
 func (ds *MySQLDatastore) Get(ctx context.Context, key []byte) ([]byte, error) {
 	row := ds.db.QueryRow("SELECT value FROM extras WHERE id=?", key)
@@ -599,7 +599,7 @@ func (ds *MySQLDatastore) Get(ctx context.Context, key []byte) ([]byte, error) {
 }
 
 /*
-Tx ...
+Tx Begins and commits a MySQL Transaction.
 */
 func (ds *MySQLDatastore) Tx(f func(*sql.Tx) error) error {
 	tx, err := ds.db.Begin()
--- a/api/datastore/mysql/mysql.go
+++ b/api/datastore/mysql/mysql.go
@@ -273,7 +273,9 @@ func (ds *MySQLDatastore) InsertRoute(ctx context.Context, route *models.Route) 
 	err = ds.Tx(func(tx *sql.Tx) error {
 		r := tx.QueryRow(`SELECT 1 FROM apps WHERE name=?`, route.AppName)
 		if err := r.Scan(new(int)); err != nil {
-			return models.ErrAppsNotFound
+			if err == sql.ErrNoRows {
+				return models.ErrAppsNotFound
+			}
 		}
 		same, err := tx.Query(`SELECT 1 FROM routes WHERE app_name=? AND path=?`,
 			route.AppName, route.Path)
--- a/api/datastore/mysql/mysql.go
+++ b/api/datastore/mysql/mysql.go
@@ -148,13 +148,10 @@ func (ds *MySQLDatastore) UpdateApp(ctx context.Context, newapp *models.App) (*m
 
 		stmt, err := ds.db.Prepare(`UPDATE apps SET config=? WHERE name=?`)
 
-<<<<<<< HEAD
 		if err != nil {
 			return err
 		}
 
-=======
->>>>>>> Make the mysql stuff work
 		res, err := stmt.Exec(string(cbyte), app.Name)
 
 		if err != nil {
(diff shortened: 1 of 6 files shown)


jmank88:
It looks like you may have cloned off an out dated version of the postgres package.

I wonder if most of this code could be shared between the two.  Are there any other differences besides the table schema?  It would be nice to have a base sql package with specific behavior overridden or plugged in.

martinpinto:
That is a good idea, since there aren't differences besides the sql schema. I will start refactoring the code and do a new commit.

denismakogon:
@martinpinto according to test logs, there's problem accessing mysql inside test environment in CircleCI. So, in order to fix that, it is recommended:
At this point, you’ll want to create your database, load it with your schema, and (possibly) preload it with data. If you use MySQL or Postgres, you can use the `circle_test` database and the `ubuntu` user rather than creating the database yourself. No password is required for any database.

See https://circleci.com/docs/1.0/manually/#databases

api/datastore/mysql/mysql_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"net/url"
+	"os/exec"
+	"testing"
+	"time"
+
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/iron-io/functions/api/datastore/internal/datastoretest"
+)
+
+const tmpMysql = "root:root@tcp(localhost:3307)/funcs"
```
denismakogon: "root:root@tcp(%v:3307)/funcs"

api/datastore/mysql/mysql_test.go
```diff
@@ -0,0 +1,106 @@
...
+	tryRun(logf, "remove old mysql container", exec.Command("docker", "rm", "-f", "iron-mysql-test"))
+	mustRun(fatalf, "start mysql container", exec.Command(
+		"docker", "run", "--name", "iron-mysql-test", "-p", "3307:3306", "-e", "MYSQL_DATABASE=funcs",
+		"-e", "MYSQL_ROOT_PASSWORD=root", "-d", "mysql"))
+	maxWait := 16 * time.Second
+	wait := 2 * time.Second
+	var db *sql.DB
+	var err error
+	for {
+		db, err = sql.Open("mysql", "root:root@tcp(localhost:3307)/")
```
denismakogon: `fmt.Sprintf("root:root@tcp(%v:3307)/", datastoretest.GetContainerHostIP())`

api/datastore/mysql/mysql_test.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	_, err = db.Exec(`GRANT ALL PRIVILEGES ON funcs.* TO root@localhost WITH GRANT OPTION;`)
+	if err != nil {
+		fmt.Println("failed to grant priviledges to user 'mysql:", err)
+		panic(err)
+	}
+
+	fmt.Println("mysql for test ready")
+	return func() {
+			db, err := sql.Open("mysql", "root:root@tcp(localhost:13307)/")
```
denismakogon: `fmt.Sprintf("root:root@tcp(%v:3307)/", datastoretest.GetContainerHostIP())`

api/datastore/mysql/mysql_test.go
```diff
@@ -0,0 +1,106 @@
...
+		func() {
+			tryRun(logf, "stop mysql container", exec.Command("docker", "rm", "-f", "iron-mysql-test"))
+		}
+}
+
+func TestDatastore(t *testing.T) {
+	_, close := prepareMysqlTest(t.Logf, t.Fatalf)
+	defer close()
+
+	u, err := url.Parse(tmpMysql)
```
denismakogon: `fmt.Sprintf(tmpMysql, datastoretest.GetContainerHostIP())`

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+	"database/sql"
+	"encoding/json"
+	"fmt"
+	"net/url"
+
+	"github.com/Sirupsen/logrus"
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/iron-io/functions/api/datastore/internal/datastoreutil"
+	"github.com/iron-io/functions/api/models"
+	_ "github.com/lib/pq"
```
jmank88: Extra import

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+	var cbyte []byte
+	var err error
+
+	if app == nil {
+		return nil, models.ErrDatastoreEmptyApp
+	}
+
+	if app.Name == "" {
+		return nil, models.ErrDatastoreEmptyAppName
+	}
```
jmank88: Redundant nil/empty checks. #579 removed these from the `postgres` package, and most of it's other changes still apply here too.

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+
+	if app.Config != nil {
+		cbyte, err = json.Marshal(app.Config)
+		if err != nil {
+			return nil, err
+		}
+	}
+	stmt, err := ds.db.Prepare("INSERT apps SET name=?,config=?")
```
jmank88: Unchecked error

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+
+	_, err = stmt.Exec(app.Name, string(cbyte))
+
+	if err != nil {
+		return nil, models.ErrAppsAlreadyExists
```
jmank88: Not necessarily?
Is there a mysql specific error that can be checked for like with postgres?

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+		}
+
+		app.UpdateConfig(newapp.Config)
+
+		cbyte, err := json.Marshal(app.Config)
+		if err != nil {
+			return err
+		}
+
+		stmt, err := ds.db.Prepare(`UPDATE apps SET config=? WHERE name=?`)
```
jmank88: Unchecked error

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+	return err
+}
+
+/*
+GetApps ...
+*/
+func (ds *MySQLDatastore) GetApps(ctx context.Context, filter *models.AppFilter) ([]*models.App, error) {
+	res := []*models.App{}
+	filterQuery, args := buildFilterAppQuery(filter)
+	rows, err := ds.db.Query(fmt.Sprintf("SELECT DISTINCT * FROM apps %s", filterQuery), args...)
```
jmank88: Coding a `SELECT *` seems like risky business

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+		return nil, err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var app models.App
+		err := scanApp(rows, &app)
+
+		if err != nil {
+			if err == sql.ErrNoRows {
```
jmank88: This looks strange to me.  Does it make sense for `rows.Next()` to return true and continue the loop if scan is going to return `sql.ErrNoRows`?  I would expect the loop to break, and this error to be returned from `rows.Err()`.
The docs actually say `ErrNoRows is returned by Scan when QueryRow doesn't return a row`, suggesting this error will never be returned from `Query()` anyways.

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+
+	cbyte, err := json.Marshal(route.Config)
+	if err != nil {
+		return nil, err
+	}
+
+	err = ds.Tx(func(tx *sql.Tx) error {
+		r := tx.QueryRow(`SELECT 1 FROM apps WHERE name=?`, route.AppName)
+		if err := r.Scan(new(int)); err != nil {
+			return models.ErrAppsNotFound
```
jmank88: Not necessarily?

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+		)
+
+		if err != nil {
+			return err
+		}
+
+		if n, err := res.RowsAffected(); err != nil {
+			return err
+		} else if n == 0 {
+			return models.ErrRoutesNotFound
```
jmank88: `ErrRoutesNotFound` should be caught above, so affecting 0 rows might be worth logging here since it would indicate something pretty severely wrong like a bug or DB integrity problems.

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var route models.Route
+		err := scanRoute(rows, &route)
+		if err != nil {
+			continue
```
jmank88: I mentioned this elsewhere, but it's worth pointing out in context: `scanRoute()` errors are tossed here.  Although it might make sense to return a best effort to the users, the error is probably worth logging since it's likely bad data or buggy scanning.

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
```
jmank88: Same as above

api/datastore/mysql/mysql_test.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"bytes"
+	"database/sql"
+	"fmt"
+	"net/url"
+	"os/exec"
+	"testing"
+	"time"
+
+	_ "github.com/go-sql-driver/mysql"
```
jmank88: Redundant import?

api/datastore/mysql/mysql.go
martinpinto: Done, removed it

api/datastore/mysql/mysql.go
martinpinto: Done, removed these to stay consistent with the postgres package.

api/datastore/mysql/mysql.go
martinpinto: Done, added check.

api/datastore/mysql/mysql.go
martinpinto: Done, added check.

api/datastore/mysql/mysql.go
martinpinto: what do you suggest how we could improve this?

api/datastore/mysql/mysql.go
martinpinto: You're right, it doesn't make much sense. Added a break and the error will be returned by `rows.Err()`.

api/datastore/mysql/mysql.go
martinpinto: well, the MySQL exception I get is `sql: no rows in result set`, which imho translates to `ErrAppsNotFound`.

api/datastore/mysql/mysql.go
martinpinto: Done, `ErrRoutesNotFound` was caught above and added logging.

api/datastore/mysql/mysql.go
martinpinto: Done, removed continue and added logging.

api/datastore/mysql/mysql_test.go
martinpinto: Done, removed import.

api/datastore/mysql/mysql.go
jmank88: Oh I just meant to include the literal column names in the query, so that `scanApp` doesn't depend on the set or order of the columns returned by `*`.

api/datastore/mysql/mysql.go
jmank88: That sounds like `sql.ErrNoRows`. Can that be checked for explicitly, and the actual err returned otherwise?

api/datastore/mysql/mysql.go
```diff
@@ -511,7 +505,7 @@ func (ds *MySQLDatastore) GetRoutesByApp(ctx context.Context, appName string, fi
 		var route models.Route
 		err := scanRoute(rows, &route)
 		if err != nil {
-			continue
+			fmt.Println("failed to get routes by app:", err)
```
jmank88: I think these still need to continue also, so the errored route is not appended.

api/datastore/mysql/mysql.go
```diff
@@ -376,9 +368,11 @@ func (ds *MySQLDatastore) UpdateRoute(ctx context.Context, newroute *models.Rout
...
 
 		if n, err := res.RowsAffected(); err != nil {
+			if n == 0 {
+				fmt.Printf("No rows affected: %s", err)
+				return models.ErrRoutesNotFound
+			}
 			return err
-		} else if n == 0 {
-			return models.ErrRoutesNotFound
 		}
```
jmank88: Won't this miss the case where `res.RowsAffected()` returns `0, nil` now?

api/datastore/mysql/mysql.go
```diff
@@ -246,10 +241,7 @@ func (ds *MySQLDatastore) GetApps(ctx context.Context, filter *models.AppFilter)
 		err := scanApp(rows, &app)
 
 		if err != nil {
-			if err == sql.ErrNoRows {
-				return res, nil
-			}
-			return res, err
+			break
```
jmank88: Can this log and continue, just like with routes?

api/datastore/mysql/mysql.go
martinpinto: oh yes of course! Changed `*` with `name, config`. Thank you!

api/datastore/mysql/mysql.go
martinpinto: Added check for `sql.ErrNoRows`.

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return nil, models.ErrAppsNotFound
+		}
+		return nil, err
+	}
+
+	return res, nil
+}
```
jmank88: Some more of the changes from #579 still apply to the above and below methods.

seiflotfy:
I think CLA is broken on this one

api/datastore/mysql/mysql.go
jmank88: Was this reply meant for another comment?
Original comment still applies.

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+			return err
+		}
+
+		if n, err := res.RowsAffected(); err != nil {
+			if n == 0 {
+				fmt.Printf("No rows affected: %s", err)
+				return models.ErrRoutesNotFound
+			}
+			return err
+		}
```
jmank88: This no longer handles the `0, nil` case.

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+
+	if headerStr == "" {
+		return models.ErrRoutesNotFound
+	}
+
+	json.Unmarshal([]byte(headerStr), &route.Headers)
+	json.Unmarshal([]byte(configStr), &route.Config)
+
+	return err
+}
```
jmank88: More changes from #579 apply here.

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var route models.Route
+		err := scanRoute(rows, &route)
+		if err != nil {
+			fmt.Println("failed to get routes:", err)
```
jmank88: This should still `continue` after logging, to skip the append (and probably use `logrus`?)

api/datastore/mysql/mysql.go
```diff
@@ -0,0 +1,615 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var route models.Route
+		err := scanRoute(rows, &route)
+		if err != nil {
+			fmt.Println("failed to get routes by app:", err)
```
jmank88: Same as above

treeder:
Need docs for this: #596

PR #576, opened 2017-03-09
Adding ability to use private registries

mkubarycz:


seiflotfy:
testing it in a bit

treeder:
This looks great!  But looks like there may be some overlap with #547, which adds an `fn docker login` command. I have a feeling these can coexist as yours appears to take env vars whereas #547 stores the auth in the database (which may make more sense when we have support for authenticating different users and they use their own registry credentials). 

Also, could you add some docs for this?

seiflotfy:
@mkubarycz could you add some docs?

api/runner/task.go
```diff
@@ -2,10 +2,12 @@ package runner
```
denismakogon: Shouldn't this code be at iron-io/runner rather than in functions? All this code is driver-specific.

pedronasser:
@mkubarycz I've taken the freedom of sending a PR to your fork, adding something. Please check it.
Beside that IMO this PR looks good.

I agree the ideal state would have been adding this to the iron/runner. But since we currently only supporting Docker and the code related to it is quite modular, in the future we can separate it from this repo.

I think this PR is a start towards supporting private registries and this can be improved in the future

cc @treeder @seiflotfy

pedronasser:
Making a new PR to continue this PR

Issue #567, opened 2017-03-03
Host header not available as env var HEADER_HOST for function to read.


jconning:
This documentation page says that the environment variable HEADER_X will be available to the function to read, for each HTTP header that was passed with the request.  
https://github.com/iron-io/functions/blob/master/docs/writing.md#inputs

In fact, HEADER_ACCEPT and HEADER_USER_AGENT are available and match the input.

However, when header "Host" is passed in, the env var HEADER_HOST is not available to the function.  Curl passes the header "Host" by default, and it would be good to have HEADER_HOST available as a means of detecting which host was called by the client.

The following curl output shows the "Host" header being passed in and also lists the environment variables available to the function (os.Environ()).

```
Jims-MacBook-Pro:eratosthenes-go jimc$ curl -v http://54.90.201.107/r/myapp/myroute
*   Trying 54.90.201.107...
* TCP_NODELAY set
* Connected to 54.90.201.107 (54.90.201.107) port 80 (#0)
> GET /r/myapp/myroute HTTP/1.1
> Host: 54.90.201.107
> User-Agent: curl/7.51.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Fri, 03 Mar 2017 21:23:19 GMT
< Content-Length: 320
< Content-Type: text/plain; charset=utf-8
<
In function now, os.Environ(): [PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin HOSTNAME=3ff3ea1030af REQUEST_URL=/r/myapp/myroute HEADER_ACCEPT=*/* HEADER_USER_AGENT=curl/7.51.0 METHOD=GET ROUTE=/myroute HOME=/root]
* Curl_http_done: called premature == 0
* Connection #0 to host 54.90.201.107 left intact
```

denismakogon:
@jconning it appears that https://github.com/iron-io/functions/blob/fae66764b4d34d3b15bb4688dc691637375df7ce/api/server/runner.go#L184 Request.Header from Gin context does not include Host header, that's why it's not available on function side as `HEADER_HOST`.

denismakogon:
Submitting patch. Tested manually with `denismakogon/os.environ` image. See execution logs:
```
curl -v -X POST localhost:8080/r/testapp/env
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> POST /r/testapp/env HTTP/1.1
[...]

PR #578, opened 2017-03-09
Add Host header to function environ as HEADER_HOST

denismakogon:
From godoc:
	For incoming requests, the Host header is promoted to the
	Request.Host field and removed from the Header map.

 So, Host header is available at c.Request.Host.

 Fixes: #567


jconning:
Good catch, @denismakogon.

PR #579, opened 2017-03-10
Postgres datastore errors and cleanup

jmank88:
This change picks a few small things from #523, and improves some additional error handling.

@martinpinto If it would be smoother to wait for #575 to merge first, or to open a PR against that branch, let me know.

### DIFF
--- a/api/datastore/postgres/postgres.go
+++ b/api/datastore/postgres/postgres.go
@@ -47,10 +47,6 @@ type rowScanner interface {
 	Scan(dest ...interface{}) error
 }
 
-type rowQuerier interface {
-	QueryRow(query string, args ...interface{}) *sql.Row
-}
-
 type PostgresDatastore struct {
 	db *sql.DB
 }
@@ -177,17 +173,18 @@ func (ds *PostgresDatastore) GetApp(ctx context.Context, name string) (*models.A
 	var resName string
 	var config string
 	err := row.Scan(&resName, &config)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return nil, models.ErrAppsNotFound
+		}
+		return nil, err
+	}
 
 	res := &models.App{
 		Name: resName,
 	}
 
-	json.Unmarshal([]byte(config), &res.Config)
-
-	if err != nil {
-		if err == sql.ErrNoRows {
-			return nil, models.ErrAppsNotFound
-		}
+	if err := json.Unmarshal([]byte(config), &res.Config); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


api/datastore/postgres/postgres.go
```diff
@@ -47,10 +47,6 @@ type rowScanner interface {
 	Scan(dest ...interface{}) error
 }
 
-type rowQuerier interface {
-	QueryRow(query string, args ...interface{}) *sql.Row
-}
-
```
jmank88: Unused

api/datastore/postgres/postgres.go
```diff
@@ -88,14 +84,6 @@ func (ds *PostgresDatastore) InsertApp(ctx context.Context, app *models.App) (*m
...
 	var err error
 
-	if app == nil {
-		return nil, models.ErrDatastoreEmptyApp
-	}
-
-	if app.Name == "" {
-		return nil, models.ErrDatastoreEmptyAppName
-	}
-
```
jmank88: Redundant

api/datastore/postgres/postgres.go
```diff
@@ -411,15 +401,18 @@ func scanRoute(scanner rowScanner, route *models.Route) error {
 		&headerStr,
 		&configStr,
 	)
+	if err != nil {
+		return err
+	}
 
 	if headerStr == "" {
 		return models.ErrRoutesNotFound
 	}
```
jmank88: Can anyone explain this one?  Why must the `headerStr` field be populated?  Why is this seemingly bad-data reported the same as a route not being found?  Note that it is not only called from `GetRoute` but `GetRoutes` as well, which does some questionable error discarding.

jmank88: Additionally, this `headerStr` check was happening *before* checking the `Scan` err.

api/datastore/postgres/postgres.go
```diff
@@ -185,17 +173,18 @@ func (ds *PostgresDatastore) GetApp(ctx context.Context, name string) (*models.A
 	var resName string
 	var config string
 	err := row.Scan(&resName, &config)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return nil, models.ErrAppsNotFound
+		}
+		return nil, err
+	}
```
denismakogon: Shouldn't we include some logging here?

api/datastore/postgres/postgres.go
```diff
@@ -209,10 +198,11 @@ func scanApp(scanner rowScanner, app *models.App) error {
 		&app.Name,
 		&configStr,
 	)
+	if err != nil {
+		return err
```
denismakogon: same here, it would be useful at least at DEBUG level.

api/datastore/postgres/postgres.go
jmank88: I haven't looked everywhere, but generally the caller of `Datastore` methods handles the logging (with `api/server/error_response.go` or something similar).

jmank88: I would like to wrap most of the errors (the random, non-static ones) across the Datsatore with more context, but I didn't want to break local style too much.

PR #580, opened 2017-03-10
Fix REQUEST_URL to reflect host and port of a client

denismakogon:
Fixes: #578

### DIFF
--- a/api/server/runner.go
+++ b/api/server/runner.go
@@ -164,7 +164,11 @@ func (s *Server) serve(ctx context.Context, c *gin.Context, appName string, foun
 	envVars := map[string]string{
 		"METHOD":      c.Request.Method,
 		"ROUTE":       found.Path,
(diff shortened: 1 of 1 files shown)


api/server/runner.go
```diff
@@ -164,7 +165,7 @@ func (s *Server) serve(ctx context.Context, c *gin.Context, appName string, foun
 	envVars := map[string]string{
 		"METHOD":      c.Request.Method,
 		"ROUTE":       found.Path,
-		"REQUEST_URL": c.Request.URL.String(),
+		"REQUEST_URL": fmt.Sprintf("%v%v", viper.GetString(EnvAPIURL), c.Request.URL.String()),
```
treeder: Using the `API_URL` in this change is not the correct URL. But why is `c.Request.URL.String()` not returning the full URL?  Might have to use Request.Host or Request.RequestURI or something.

PR #581, opened 2017-03-11
Mq tests

### LINKED ISSUE #574: missing tests for MQs
There are no tests for the MQs, this needs to be changed to new MQs can be written easier.

jmank88:
This change adds tests to `api/mqs` for `models.MessageQueue`s (fixes #574).  Some implementations were fixed/tweaked to pass the tests, and there is just one major outstanding inconsistency between them.  Only the bolt implementation is preserving the original push order when reservations time out.  The others appear to be re-queuing at the back of the queue (and failing the test `fifo-timeout`). The doc from the interface is:
```
// When a job is required to be restored to the queue, it should maintain it's
// approximate order in the queue. That is, for jobs [A, B, C], with A being
// the head of the queue:
// Reserve() leads to A being passed to a consumer, and timeout started.
// Next Reserve() leads to B being dequeued. This consumer finishes running the
// task, leading to Delete() being called. B is now permanently erased from the
// queue.
// A's timeout occurs before the job is finished. At this point the ordering
// should be [A, C] and not [C, A].
```  
Should this behavior be fixed?  Or can the test be relaxed to accept any sort of re-queuing?


api/mqs/mqs_test.go
```diff
@@ -0,0 +1,630 @@
...
+		if got, err := mq.Reserve(ctx); err != nil {
+			t.Log(logBuf.String())
+			t.Fatalf("unexpected error on reserve: `%v`", err)
+		} else if !reflect.DeepEqual(*task, *got) {
+			t.Log(logBuf.String())
+			t.Fatalf("pushed %v but reserved %v", task, got)
+		}
+
+		// Let reservation timeout
+		<-time.After(20 * testReserveTimeout)
```
jmank88: Even with `testReserveTimeout` at 2 seconds (and tickers at 1 second), I had to crank up the wait time on timeouts to get all the implementations to pass.

jmank88: It occurred to me that the impls using tickers could be changed to use a manually triggered tick channel for tests, so we can force the background goroutines, rather than try to wait 'long enough'.

api/mqs/ironmq.go
```diff
@@ -68,34 +69,38 @@ func NewIronMQ(url *url.URL) *IronMQ {
 	}
 
 	var queueName string
-	if url.Path != "" {
-		queueName = url.Path
+	if len(url.Path) > 1 {
+		// Drop leading '/'.
+		queueName = url.Path[1:]
```
sheerun: I just had this error ;/

jmank88: Hopefully we can get this fix in soon. Waiting on #599 at the moment, then the rest of this PR will follow in small portions.

api/mqs/ironmq.go
```diff
@@ -37,8 +39,7 @@ type IronMQConfig struct {
 	QueuePrefix string `mapstructure:"queue_prefix"`
 }
 
-func NewIronMQ(url *url.URL) *IronMQ {
-
+func NewIronMQ(url *url.URL, reserveTimeout time.Duration) (*IronMQ, error) {
```
sheerun: There's no `time` import?

jmank88: This PR is on hold in an inconsistent state while smaller pieces are merged in one at a time, starting with #599. This PR will narrow in scope and be cleaned up later, or closed in favor of another.

PR #597, opened 2017-03-21
Add env vars for hotfunctions

pedronasser:
So functions can know if they are running as hot function or not and which format has been defined.


api/runner/worker.go
```diff
@@ -296,6 +296,8 @@ func (hc *htfn) serve(ctx context.Context) {
 	}()
 
 	cfg := *hc.cfg
+	cfg.Env["FN_HOT"] = "true"
```
treeder: I think hot is assumed if format is not `default` so probably don't need the extra var.

PR #599, opened 2017-03-22
adding generic mqs tests; testing memory impl

jmank88:
This change pulls some of the tests from #581 (towards #574), and runs them against the memory implementation.  Future PRs will test other implementations, and add more tests.


ucirello (commented):
LGTM+1 but I'd like to have more eye balls on this change. @seiflotfy @pedronasser

jmank88:
Fixed the CI on this one

PR #601, opened 2017-03-22
Added Lambda Node support as part of the regular functions workflow.

treeder:
Using `--runtime lambda-node` will use the lambda wrapper for the function. 

```sh
fn init --runtime lambda-node <DOCKER_HUB_USERNAME>/lambda-node
```

The PR looks a lot bigger than it is, a lot of the files that were changed are minor. The main thing is the `fn/lambda` stuff that builds the `iron/functions-lambda:node` image. And I've added a `lambda-node` runtime that will use that image. Lastly, added "cmd" value to `func.yaml` since the lambda image requires use of the `ENTRYPOINT`.

### DIFF
--- a/fn/common.go
+++ b/fn/common.go
@@ -162,6 +162,7 @@ func writeTmpDockerfile(dir string, ff *funcfile) error {
 	if err != nil {
 		return err
 	}
+	defer fd.Close()
 
 	// convert entrypoint string to slice
 	bufferEp := stringToSlice(ff.Entrypoint)
@@ -171,7 +172,7 @@ func writeTmpDockerfile(dir string, ff *funcfile) error {
 	err = t.Execute(fd, struct {
 		BaseImage, Entrypoint, Cmd string
 	}{rt, bufferEp.String(), bufferCmd.String()})
-	fd.Close()
+
 	return err
 }
 
--- a/fn/common.go
+++ b/fn/common.go
@@ -79,9 +79,9 @@ func dockerbuild(verbwriter io.Writer, path string, ff *funcfile) error {
 		if err != nil {
 			return err
 		}
-		helper, err = langs.GetLangHelper(*ff.Runtime)
-		if err != nil {
-			return err
+		helper = langs.GetLangHelper(*ff.Runtime)
+		if helper == nil {
+			return fmt.Errorf("Cannot build, no language helper found for %v", *ff.Runtime)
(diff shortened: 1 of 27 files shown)


treeder:
cc @jmank88

fn/common.go
```diff
@@ -162,7 +164,19 @@ func writeTmpDockerfile(dir string, ff *funcfile) error {
...
 	// convert entrypoint string to slice
-	epvals := strings.Fields(*ff.Entrypoint)
+	bufferEp := stringToSlice(ff.Entrypoint)
+	bufferCmd := stringToSlice(ff.Cmd)
+
+	t := template.Must(template.New("Dockerfile").Parse(tplDockerfile))
+	err = t.Execute(fd, struct {
+		BaseImage, Entrypoint, Cmd string
+	}{rt, bufferEp.String(), bufferCmd.String()})
+	fd.Close()
```
jmank88: Should this `Close()` call be deferred up around ~165?  Though it not matter in the big picture, since something like a `Must()` `panic` would terminate the process anyways.

fn/init.go
```diff
@@ -157,16 +159,29 @@ func (a *initFnCmd) buildFuncFile(c *cli.Context) error {
...
+		fmt.Printf("No helper found for %s runtime, you'll have to pass in the appropriate flags or use a Dockerfile.", a.runtime)
+	}
+
 	if a.entrypoint == "" {
-		ep, err := detectEntrypoint(a.runtime)
-		if err != nil {
-			return fmt.Errorf("could not detect entrypoint for %v, use --entrypoint to add it explicitly. %v", a.runtime, err)
+		if helper != nil {
+			a.entrypoint = helper.Entrypoint()
+			// return fmt.Errorf("could not detect entrypoint for %v, use --entrypoint to add it explicitly. %v", a.runtime, err)
```
jmank88: Remove?

seiflotfy:
There lambda import support is missing I am working on that now

fn/common.go
treeder: Makes sense, changing.

treeder:
I think we can ditch all the other lambda related commands (except import) as long as these special runtimes will cover it using the regular functions workflow.

fn/init.go
treeder: Removed.

Issue #554, opened 2017-03-01
Hot function inactivity timeout


WTFKr0:
Hi
What do you think about making the 30 seconds inactivity timeout a parameter for hot functions (http mode) ?
I can think about a function with high overhead in loading or database connection
If this function is called every 2 min from 8am to 9am for example, I think it would be great to tune the timeout efficiently
WDYT ?

pedronasser:
@WTFKr0, when you say parameter do you mean a route configuration?

I agree in making the hot functions timeout configurable.

WTFKr0:
Yep route configuration

denismakogon:
@WTFKr0 I assume there's only one way to define timeout:

- extend HTTP Post to `/{app}/routes` to include new parameter (let's say, `inactivity_timeout`), but make it optional (similar to other optional parameter we maintain right now)

Would that make sense @pedronasser @seiflotfy @treeder ?

PR #603, opened 2017-03-22
Add idle_timeout to routes API

denismakogon:
Adding `inactivity_timeout` to routes API. 
This change will allow users to define idle timeout for hot functions via API instead of having it hardcoded in `worker.go`.

Closes: #554

### DIFF
--- a/api/models/new_task.go
+++ b/api/models/new_task.go
@@ -54,6 +54,11 @@ type NewTask struct {
 
 	*/
 	Timeout *int32 `json:"timeout,omitempty"`
+
+	/* Hot function inactivity timeout in seconds before termination.
+
+	*/
+	InactivityTimeout *int32 `json:"inactivity_timeout,omitempty"`
 }
 
 // Validate validates this new task
--- a/api/models/new_task.go
+++ b/api/models/new_task.go
@@ -55,10 +55,10 @@ type NewTask struct {
 	*/
 	Timeout *int32 `json:"timeout,omitempty"`
 
-	/* Hot function inactivity timeout in seconds before termination.
+	/* Hot function idle timeout in seconds before termination.
 
 	*/
-	InactivityTimeout *int32 `json:"inactivity_timeout,omitempty"`
+	IdleTimeout *int32 `json:"inactivity_timeout,omitempty"`
 }
 
 // Validate validates this new task
--- a/api/models/new_task.go
+++ b/api/models/new_task.go
@@ -58,7 +58,7 @@ type NewTask struct {
 	/* Hot function idle timeout in seconds before termination.
 
 	*/
-	IdleTimeout *int32 `json:"inactivity_timeout,omitempty"`
+	IdleTimeout *int32 `json:"idle_timeout,omitempty"`
 }
 
 // Validate validates this new task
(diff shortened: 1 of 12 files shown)


treeder:
Generally looks good, but can we call it `IdleTimeout` instead of `InactivityTimeout` ?

denismakogon:
@treeder Sure, why not.

api/models/new_task.go
```diff
@@ -54,6 +54,11 @@ type NewTask struct {
 
 	*/
 	Timeout *int32 `json:"timeout,omitempty"`
+
+	/* Hot function idle timeout in seconds before termination.
+
+	*/
+	IdleTimeout *int32 `json:"inactivity_timeout,omitempty"`
```
jmank88: There are a few dangling 'inactivity' usages in json fields and errors messages

denismakogon:
@ccirello So, yes, there are two timeouts, and it doesn't look like something bad or confusing. One timeout is for function call, and the idle timeout defines for how long to keep hot function around without serving requests for it. You're saying that it was not thought through, i tend to disagree on this, because it's not clear why hardcoded 30s timeout is exact timeout that every hot function should wait without processing requests.
From user perspective, if i have a workload that comes each 1m, and i don't want to spend time to wait until function will be started (+ overhead of function itself, like pulling out data from 3rd party services, etc.) rather than setting idle timeout for 2m, so with such workload my hot function will be more effective in terms of faster requests serving.
Does it make sense?
cc @treeder @seiflotfy @pedronasser

ccirello:
@denismakogon - makes sense to a point.

> So, yes, there are two timeouts, and it doesn't look like something bad or confusing. 

Without context, I doubt people will know the difference between timeout and idle_timeout. This could be easily addressed with documentation though - as of now, however, docs are missing.

> You're saying that it was not thought through, i tend to disagree on this, because it's not clear why hardcoded 30s timeout is exact timeout that every hot function should wait without processing requests before becoming "cold".

Indeed, it is not clear the reason behind 30s. However, whenever we change a constant to a variable, we are changing the behavior of the system. So what you are saying is that the user needs it - which I am fine with - and that this reason alone is enough to justify this change as it is right now.

I did not see anywhere any consideration of the consequences of this change: it does not define min and max boundaries, does not explain the consequence of too short or too long idle timeouts, it seems to address the zero case implicitly (for instance, it is not self-evident for me what happens if this idle_timeout is zero); also I do not see in the source code the interaction between this field being updated in the DB while the hot functions are still running. 

Again, I am not against the goal of this change - but this implementation of this change is lacking in many aspects. 

A middle-way here would be to merge this as it is, and open a batch of tickets addresses its deficiencies.

denismakogon:
@ccirello I agree on documentation as well as with edge cases. I'll make next PR with docs that explicitly describes all possible cases for idle time and its combinations with regular timeout.

treeder:
The docs should be part of this PR, agree with @ccirello . Not complete without docs.

denismakogon:
@treeder doc is already here. Check last 2 commits.

treeder:
@ccirello ok now with docs?

ccirello:
@denismakogon @treeder LGTM

PR #604, opened 2017-03-22
map nodejs4.3 to nodejs as base

seiflotfy:


seiflotfy:
this is not needed anymore...

PR #605, opened 2017-03-23
Fix lambda import

seiflotfy:


fn/lambda.go
```diff
@@ -214,9 +110,13 @@ func awsImport(c *cli.Context) error {
 		return err
 	}
 
+	if *function.Configuration.Runtime == "nodejs4.3" {
+		*function.Configuration.Runtime = "node"
```
treeder: It thought we decided to just keep the lambda names?

fn/lambda.go
```diff
@@ -214,9 +110,13 @@ func awsImport(c *cli.Context) error {
...
+	}
+
 	opts := createImageOptions{
 		Name:          functionName,
-		Base:          fmt.Sprintf("iron/lambda-%s", *function.Configuration.Runtime),
+		Base:          fmt.Sprintf("lambda-%s", *function.Configuration.Runtime),
```
treeder: `functions/lambda-%s` ?

fn/lambda.go
```diff
@@ -239,12 +139,18 @@ func awsImport(c *cli.Context) error {
...
+	}
+	fmt.Println("OK")
+
+	return nil
 }
 
 var (
 	runtimeImportHandlers = map[string]func(functionName, tmpFileName string, opts *createImageOptions) ([]fileLike, error){
-		"nodejs":    basicImportHandler,
+		"node":      basicImportHandler,
```
treeder: node4.3?

fn/lambda.go
seiflotfy: that would require updating the names of the docker images first for this PR to work

fn/lambda.go
seiflotfy: the init produces
```
runtime: lambda-node
```

This serves the same purpose

fn/lambda.go
seiflotfy: Again requires updating the image names for this to work

fn/lambda.go
treeder: I don't even think this will pull any image though regardless, since you don't even have a docker hub username anymore. Let's assume we'll push the images with the following naming: `iron/functions-lambda:{aws_name}`

PR #610, opened 2017-03-28
Solving postgres marshal/unmarshal issue

pedronasser:
Postgres datastore was not marshaling the App config during its insert, that behavior was resulting in issues when fetching the App and the datastore couldn't unmarshal the config.

The same issue was probably happening with the Route's headers in some situations.

This commit's idea is to always try to marshal configs and headers when inserting/updating Apps or Routes. But in Apps and Routes get methods, if the  config/headers unmarshal fails, it returns an empty config/headers.

### DIFF
--- a/api/datastore/postgres/postgres.go
+++ b/api/datastore/postgres/postgres.go
@@ -122,7 +122,10 @@ func (ds *PostgresDatastore) UpdateApp(ctx context.Context, newapp *models.App) 
 		}
 
 		if len(config) > 0 {
-			json.Unmarshal([]byte(config), &app.Config)
+			err := json.Unmarshal([]byte(config), &app.Config)
+			if err != nil {
+				return err
+			}
 		}
 
 		app.UpdateConfig(newapp.Config)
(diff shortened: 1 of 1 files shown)


api/datastore/postgres/postgres.go
```diff
@@ -410,14 +411,15 @@ func scanRoute(scanner rowScanner, route *models.Route) error {
...
-		return models.ErrRoutesNotFound
+	if len(headerStr) > 0 {
+		json.Unmarshal([]byte(headerStr), &route.Headers)
 	}
 
-	if err := json.Unmarshal([]byte(headerStr), &route.Headers); err != nil {
-		return err
+	if len(configStr) > 0 {
+		json.Unmarshal([]byte(configStr), &route.Config)
 	}
```
jmank88: I understand bypassing these if they are nil/empty, but why ignore the returned errors?

pedronasser: Agreed, I'm changing it to return errors when unmarshaling non-empty strings.

PR #625, opened 2017-04-19
fn: support https API url

pedronasser:
Add support to https API's URL in fn

### DIFF
--- a/fn/api.go
+++ b/fn/api.go
@@ -33,7 +33,9 @@ func scheme() string {
 }
 
 func apiClient() *fnclient.Functions {
-	transport := httptransport.New(host(), "/v1", []string{scheme()})
+	url := apiURL()
+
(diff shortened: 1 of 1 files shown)


fn/api.go
```diff
@@ -21,11 +21,20 @@ func host() string {
...
+	return apiURL().Host
+}
+
+func scheme() string {
+	return apiURL().Scheme
 }
 
 func apiClient() *fnclient.Functions {
-	transport := httptransport.New(host(), "/v1", []string{"http"})
+	transport := httptransport.New(host(), "/v1", []string{scheme()})
```
jmank88: It seems a little redundant and racy to get and parse the url twice in a row. What about calling `apiURL` once from here and accessing the `Host` and `Scheme` fields directly?

Issue #594, opened 2017-03-20
Unable to build image


james5233:
The following error comes up when trying to build the 'hello' example following the steps detailed in the README:
$ fn build
Step 3 : ADD functions-alpine /app/functions
lstat functions-alpine: no such file or directory
error running docker build: exit status 1

treeder:
Are you trying to build int the root dir of the `iron-io/functions` repo? Make a fresh new directory and with nothing in it and start there.

PR #629, opened 2017-07-12
Hopefully help fill in an "obvious" step or two

alnutile:
I ran into the same problem as seen here https://github.com/iron-io/functions/issues/594

The path issue is different but maybe that will help too.


c0ze:
LGTM

alnutile:
thanks!

Issue #639, opened 2017-08-18
fn push needs to also tag latest


prologic:
I noticed when doing an `fn push` it pushes and only tags the version specified in the `func.yml`.
This is unideal as its an extra step and extra version you have to add routes for.

Can we also tag `latest` to whatever is pushed so latest is always pointing at the latest pushed version?

PR #646, opened 2017-09-14
[skip ci] Updated the document of examples/echo

kunihiko-t:
[skip ci] When I ran some examples on my local env, I noticed that there are errors in some commands.
So I updated the document.


examples/echo/README.md
```diff
@@ -23,7 +23,8 @@ USERNAME=YOUR_DOCKER_HUB_USERNAME
 ```
 # tagging
 docker run --rm -v "$PWD":/app treeder/bump patch
-docker tag $USERNAME/func-echo:latest $USERNAME/func-echo:`cat VERSION`
+docker tag iron/func-echo:latest $USERNAME/func-echo:`cat VERSION`
+docker tag $USERNAME/func-echo:`cat VERSION` $USERNAME/func-echo:latest
```
c0ze: Actually, fn tool now automatically tags latest, so probably this is redundant.

#639

kunihiko-t: I see.
I'll undo the changes.
Thanks.

c0ze: Thanks for the update!

Actually, now that latest tag is added automatically, we can update the docs to remove adding latest tag.

Would you mind updating to latest fn tool, confirming this behavior, and removing all reference to add latest tags from the documentation?

kunihiko-t: Yeah, I'll work on it.

You mean documents of examples should have this kind of "fn" style?
https://github.com/iron-io/functions/blob/0c3206deb5f8ebda5283662a271138a84ef8ecb1/examples/hello/go/README.md

c0ze: yes, if possible !

PR #650, opened 2017-09-28
Fix ssl

### LINKED ISSUE #649: Unable to get API URL working with SSL termination through load balancers
We've got a CNAME to our ingress load balancer in Rancher for Iron Functions. Running a valid wildcard SSL through from the ingress LB for the environment, SNI to the Stack LB, termination of SSL at the stack load balancer then redirection of traffic from 443 to 8080 in the Function Workers containers. Same set up for the UI. On the UI side everything works perfectly fine. No specification of port 4000 is required. On the API said I get an error when trying to do the same thing:

MASKED:demoapp rsuarez$ fn apps create testapp2
unexpected error: no consumer: "text/html"

screen shot 2017-09-27 at 9 09 12 am

c0ze:
This should fix #649 and #635 .

For testing you can use :

[ssl_nginx_docker.zip](https://github.com/iron-io/functions/files/1339303/ssl_nginx_docker.zip)

You can use `SSL_SKIP_VERIFY` Env variable to skip SSL checking. Testing from CLI :

```
SSL_SKIP_VERIFY=true API_URL=https://localhost go run api.go         build.go       common.go      errors.go      images.go      lambda.go routes.go      run.go  version.go apps.go        bump.go        deploy.go      funcfile.go    init.go        main.go        push.go run_others.go  testfn.go version
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -305,7 +305,7 @@
   branch = "master"
   name = "github.com/iron-io/functions_go"
   packages = [".","client","client/apps","client/routes","client/tasks","client/version","models"]
-  revision = "69e4dec8454c3c710045263c2ede76139c141146"
+  revision = "8d7dbe9fd1698e16b45374c8b7348d499124e4a3"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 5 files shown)


c0ze:
we also need https://github.com/iron-io/functions_go/pull/6 for this to work.

PR #651, opened 2017-10-02
Updated the document of examples/checker

kunihiko-t:
Updated the document.

I'm not sure why extractEnvVar func adds HEADER prefix to the env variables.

https://github.com/iron-io/functions/blob/774b569bd2a6e7c2dd374913b752771491442472/fn/run.go#L105

```
echo '{"env_vars": {"HEADER_FOO": "bar"}}' | ./fn run -e FOO=bar
=> all good

echo '{"env_vars": {"FOO": "bar"}}' | ./fn run -e FOO=bar
=> Env var FOO does not match
```


c0ze:
it's because maybe functions provides headers to the function as env vars ?

https://github.com/iron-io/functions#passing-data-into-a-function

So that function can distinguish which env var comes from the request.

kunihiko-t:
@c0ze Thanks 👍

c0ze:
Thanks !

PR #657, opened 2017-11-07
refactor fn to seperate common and commands

c0ze:
@kunihiko-t I refactored fn tool to make it easier to work with. Would you like to have a look ?

With this refactor, you can just 

```
go run main.go
```

and run fn tool.


kunihiko-t:
Thanks! Let me check it out.

fn/commands/deploy.go
```diff
@@ -0,0 +1,241 @@
...
+// the route can be overriden inside the functions file.
+func (p *deploycmd) deploy(path string) error {
+	fmt.Fprintln(p.verbwriter, "deploying", path)
+
+	funcfile, err := common.Buildfunc(p.verbwriter, path)
+	if err != nil {
+		return err
+	}
+
+	if p.skippush {
```
vasilev: So if `skippush` is enabled then routes will not updated ?

fn/common/api.go
```diff
@@ -0,0 +1,28 @@
+package common
+
+import (
+	"os"
+
+	"crypto/tls"
+	httptransport "github.com/go-openapi/runtime/client"
+	"github.com/go-openapi/strfmt"
+	fnclient "github.com/iron-io/functions_go/client"
+	"net/http"
```
vasilev: Nit: Please consider to separate imports of standard packages in one group.

c0ze:
Oops, sorry actually there are some diffs I forgot to check, just a sec.

fn/commands/deploy.go
c0ze: Thanks for the review ! This part is as is with master, I didn't change any logic within functions. Yes, it seems original intention is that if you pass skippush, routes will not get updated. I am not sure if this is intentional, or a bug.

https://github.com/iron-io/functions/blob/master/fn/deploy.go#L123

fn/common/api.go
c0ze: Ok, thanks ! I'll group them in one group.

c0ze:
Sorry, I just made some changes (moved image commands into a sub folder).

This should now be ok to review. Sorry for the long diff, maybe it is better to checkout and check / test locally.

c0ze:
@kunihiko-t thank you ! I am merging this then !

If possible, can you rebase jwt_proposal to new master ?

In the mean time, I will consult KG team for ron-io/functions_go#7 .
(I can merge it, but I want to make sure it doesn't break anything for them.)

c0ze:
@kunihiko-t iron-io/functions_go#7 is merged !

kunihiko-t:
Thank you!

PR #658, opened 2017-11-08
Add image processing examples

### LINKED ISSUE #162: Image processing example
Copy one we have for IronWorker and convert it to IronFunctions. 

Great async example where the output is the destination of the data (ie: s3).

c0ze:
fixes #162

### DIFF
--- a/examples/image_processing/inplace/Dockerfile
+++ b/examples/image_processing/inplace/Dockerfile
@@ -1,7 +1,4 @@
-FROM iron/ruby
-
-RUN apk --no-cache --update add alpine-sdk gd gd-dev
-RUN apk --no-cache --update add ruby ruby-io-console ruby-irb ruby-json ruby-rake ruby-rdoc ruby-dev
+FROM iron/ruby:dev
 RUN apk --no-cache --update add imagemagick
 
 COPY ./ /func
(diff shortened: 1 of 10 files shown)


examples/image_processing/inplace/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM iron/ruby
```
vasilev: Maybe use [`iron/ruby:dev`](https://github.com/iron-io/dockers/blob/master/ruby/dev/Dockerfile) to avoid separate installs of `ruby-dev`, `bundler`, etc ?

c0ze: ah yes, that makes sense. Thank you !

c0ze:
ok, I updated the Dockerfiles. I also fixed the path at example in readme.

Btw, when I run with `fn run`, I get

```
/usr/lib/ruby/2.3.0/uri/rfc3986_parser.rb:67:in `split': bad URI(is not URI?): http://www.sourcecertain.com/img/Example.png (URI::InvalidURIError)
```

When running with a curl, there is no error, and it works perfectly. (Same URI)

Any ideas ? @vasilev @kunihiko-t  ? Thank you !

c0ze:
This should be ready to go now.

PR #660, opened 2017-11-09
Jwt proposal

kunihiko-t:
This PR enables JWT Authentication for each function


api/server/runner.go
```diff
@@ -137,6 +146,33 @@ func (s *Server) handleRequest(c *gin.Context, enqueue models.Enqueue) {
 	c.JSON(http.StatusNotFound, simpleError(models.ErrRunnerRouteNotFound))
 }
 
+func processJwt(signingKey string, c *gin.Context) error {
```
c0ze: since we are (probably) going to re use this in `server.go` when implementing JWT for base API calls, I think we should refactor this (and jwt dependencies) out of here, and into a separate package where they can be referenced from `server` and `commands`.

fn/commands/routes.go
```diff
@@ -572,3 +624,64 @@ func (a *routesCmd) delete(c *cli.Context) error {
 	fmt.Println(appName, route, "deleted")
 	return nil
 }
+
+func (a *routesCmd) token(c *cli.Context) error {
```
c0ze: this too

fn/commands/routes.go
```diff
@@ -572,3 +624,64 @@ func (a *routesCmd) delete(c *cli.Context) error {
...
+	}
+	t := struct {
+		Token string `json:"token"`
+	}{Token: ss}
+	enc.Encode(t)
+
+	return nil
+}
+
+func getJwtToken(signingKey string, expiration int) (string, error) {
```
c0ze: this too

PR #686, opened 2018-08-20
Use GitHub's URL shortener

turbo:
git.io is GitHub's own URL shortener, which is guaranteed to only point to GitHub domains. This is safer and (IMO) more professional. Additionally, goo.gl will be deprecated *very* soon.


c0ze:
thank you !

